PR #4, opened 2021-01-26
Get metav1.Condition on board

RainbowMango:
**What type of PR is this?**
/kind cleanup
/kind api-change

**What this PR does / why we need it**:
Replace `ServiceExportCondition` with `metav1.Condition` which released at v1.19.0.
cc @JeremyOT


RainbowMango:
/assign @pmorie @JeremyOT 
Kindly ping...

JeremyOT:
Thanks for making this change! This is something we definitely want, but since there are a few projects actively using this API, perhaps we should add this as v1alpha2 since it's not quite a drop-in replacement.

RainbowMango:
Yeah, that makes sense, it's a kind of API change.

> since there are a few projects actively using this API

Glad to hear that, but why not make a tagged release for them?

JeremyOT:
Tagged v0.1.0, I think this makes sense to merge. 

/lgtm
/approved

PR #11, opened 2021-09-17
Bump dependencies

howardjohn:
I was planning to make some larger changes that required this but changed my mind. But I think its still good to do this, so users are not pulling in super old versions that may have compatibility issues with their other imports.


RainbowMango:
Hi @howardjohn ,

This PR just updates some dependencies, right? Are you going to update the API further?

howardjohn:
I didn't have any plans to. It should impact folks importing the API as a
go library not users

go.mod
```diff
@@ -3,15 +3,14 @@ module sigs.k8s.io/mcs-api
...
-	sigs.k8s.io/controller-runtime v0.6.1
-	sigs.k8s.io/controller-tools v0.3.0
-	sigs.k8s.io/kind v0.8.1
+	github.com/go-logr/logr v0.4.0
+	github.com/onsi/ginkgo v1.16.4
+	github.com/onsi/gomega v1.16.0
+	k8s.io/api v0.22.2
+	k8s.io/apimachinery v0.22.2
+	k8s.io/client-go v0.22.2
+	k8s.io/code-generator v0.22.1
```
RainbowMango, line 12: I guess all of the dependencies should consistent.
```suggestion
	k8s.io/code-generator v0.22.2
```

pkg/controllers/controllers_suite_test.go
```diff
@@ -96,8 +96,9 @@ var _ = BeforeSuite(func(done Done) {
 	}
 
 	stopCh = make(chan struct{})
```
JeremyOT, line 98: stopCh needs to be replaced with a context that is closed  in AfterSuite, as is the context will be closed immediately. Did tests pass?

howardjohn: I had assumed there was some CI/CD. Are developers expected to manually run tests and report if they pass?

howardjohn: I did update the code to handle the closing in AfterSuite though.

Issue #16, opened 2022-08-08
Create Test 2 / "local service not impacted" e2e test


lauralorenz:
Per the KEP, an e2e test must exist that tests:

> Test cluster A local service not impacted by same-name imported service.

We define "not impacted" as:

> Not impacted = when you run a Service in cluster A called `foo`, and you have a ServiceImport for `foo`, you can still resolve `foo.ns.svc.cluster.local` and and its IP is different from the ServiceImport IP and when you actually request the local service IPs you get a response back and its the right response (so we have to make the response smart, such as reporting their clusterID)
> `foo.ns.svc.clusterset.local` will sometimes return clusterB
> `foo.ns.svc.cluster.local` will always (always=50 times in a row) return cluster A

Given a setup with two clusters, this test should do the following:

* Service is in ClusterA, NOT exported
    * can hit it with cluster.local
* Service is in ClusterB and exported
    * can hit it with clusterset.local, always reports ClusterB
* Also Export the Service in ClusterA 
    * can hit it with clusterset.local, sometimes will say ClusterB, sometimes will say ClusterA
* Hit it 50 times, if I get at least one from either we good
    * Can still hit it with cluster.local, will ALWAYS say its from clusterA
    * No other services should be responding
        * Relevant because if the SUT happens to have other clusters participating and they get all up in there, we want to report that

See the [existing connectivity test](https://github.com/kubernetes-sigs/mcs-api/blob/master/e2etest/connectivity_test.go) for inspiration on setup and potential request and response pods. Notably in this case the response needs to be smart enough to share what cluster it is in.

lauralorenz:
/assign lauralorenz
@knee-berts is helping me with this one

nojnhuh:
Sorry @lauralorenz didn't catch that you were already assigned to this. Don't mean to step on your toes, but happy to pick this up!

nojnhuh:
@lauralorenz Just to clarify since these tests involve setting up DNS, is compiling our own CoreDNS with https://github.com/coredns/multicluster and configuring that for kind the recommended approach? Are there any other steps required to make the `clusterset.local` domain reachable?

nojnhuh:
> @lauralorenz Just to clarify since these tests involve setting up DNS, is compiling our own CoreDNS with https://github.com/coredns/multicluster and configuring that for kind the recommended approach? Are there any other steps required to make the `clusterset.local` domain reachable?

I think I got this working here: #50

Issue #2, opened 2020-12-23
ServiceExport Client should support operations across multiple namespaces


josephpeacock:
Current Behavior:
The serviceexport client has ns as a private field in the struct (https://github.com/kubernetes-sigs/mcs-api/blob/master/pkg/client/clientset/versioned/typed/apis/v1alpha1/serviceexport.go#L55), and uses that as a hardcoded namespace value in all operations. This makes it difficult to use the client for an entire cluster, to create ServiceExport resources for Services in arbitrary namespaces.

Expected Behavior:
I can create a single client, and provide the namespace at call-time to specify CRUD operations to occur in/for a specific namespace

As a workaround, I have been directly using the RESTClient, for example:

```
func (sc *ServiceExportController) createServiceExport(service *v1.Service) (v1alpha1.ServiceExport, error) {
	serviceExport := v1alpha1.ServiceExport{}
	serviceExport.Namespace = service.Namespace
	serviceExport.Name = service.Name

	result := &v1alpha1.ServiceExport{}
	err := sc.client.Post().
		Namespace(service.Namespace).
		Resource("serviceexports").
		VersionedParams(&metav1.CreateOptions{}, scheme.ParameterCodec).
		Body(serviceExport).
		Do(context.TODO()).
		Into(result)

	return err
}
```
But that is unwieldy in the code, and makes unit testing more difficult.

josephpeacock:
This is false. I was using the wrong syntax - for anyone in the future having similar issues to me, here's how to do it right:
```
kubeClient.MCSApis().MulticlusterV1alpha1().ServiceExports("nsName").Get(context.TODO(), "serviceExportName", metav1.GetOptions{})
```

PR #32, opened 2023-01-27
e2e-test #2

knee-berts:
What type of PD is this?
/kind cleanup
/kind e2e-test

What this PD does / why we need it:
Added test from https://github.com/kubernetes-sigs/mcs-api/issues/16
Added multicluster CoreDNS plugin to test install script.


lauralorenz:
/remove-lifecycle stale
@knee-berts if you can rebase, or I can clone to rebase, I think this got trapped in a rebase circle

PR #34, opened 2023-02-10
Avoid crashing while waiting for the hello pod IP

skitt:
After listing the hello pods, check whether the list is empty before dereferencing the first one. Conceivably the "eventually" expectation could be changed to check the list itself, but we're really interested in waiting for the pod to have an IP address, so this preserves that behaviour.

Signed-off-by: Stephen Kitt <skitt@redhat.com>


tpantelis (approved):

jmhbnz (approved):
LGTM

PR #45, opened 2023-12-11
fix: conformance suite should not be with _test suffix

Xunzhuo:
conformance suite defs should not be with _test suffix, this is not accessbile for inner packages.

image


mikemorris:
@nojnhuh I know you recently got the conformance tests running, do you think a change like this is still needed?

nojnhuh:
@mikemorris I got the tests under `e2etest/` working, but haven't looked at the conformance tests yet. Are there any docs describing how the conformance tests are designed to be used?

mikemorris:
@nojnhuh Found the doc I remembered that had some detail on conformance plans https://docs.google.com/document/d/11mK-Xxug4GYFrhXmt6KdS5ApB2gAM3mf28HHbvMwqHs/edit

My recollection was that e2e tests were somewhat analogous to conformance because the MCS API intentionally doesn't include a controller, leaving the actual mechanism of watching ServiceExport resources and creating corresponding ServiceImports for vendor implementations.

nojnhuh:
Thanks @mikemorris! Regarding this particular change, I think it's necessary if we expect implementors to import these packages in go.mod and maintain the entrypoint to the tests themselves. Otherwise if it's like the [Gateway API conformance tests](https://gateway-api.sigs.k8s.io/concepts/conformance/#running-tests) where this repo will contain the entrypoint to the tests and implementors are expected to invoke that, then I don't think this change is strictly necessary.

mikemorris:
> I think it's necessary if we expect implementors to import these packages in go.mod and maintain the entrypoint to the tests themselves. Otherwise if it's like the [Gateway API conformance tests](https://gateway-api.sigs.k8s.io/concepts/conformance/#running-tests) where this repo will contain the entrypoint to the tests and implementors are expected to invoke that, then I don't think this change is strictly necessary.

That makes sense - I don't think it's a mutually exclusive decision, as Gateway API has conformance tests under a [`conformance/tests`](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance) directory without `_test` suffixes. It's definitely encouraged (and maybe required for conformance acceptance) to run the suite standalone rather than integrated into an implementation, but I know with the current maturity of some existing implementations (and the suite itself), it may be helpful to accept this change to enable some tweaks or customizations around the framework while we work towards a more mature conformance program.

@lauralorenz @skitt do y'all have thoughts on this?

lauralorenz:
It's been a while but I did expect the entrypoint to be maintained in this repo to be invoked by the implementors, but nothing is final, so I think this is fine. Reading back this in particular originates from some [ginkgo autogeneration](https://onsi.github.io/ginkgo/#mental-model-go-modules-packages-and-tests) rules that don't really apply here anyways

/lgtm
/approve

skitt:
Oops, yes, I have thoughts on this, but they align with Laura’s — I do think there’s value in providing a conformance suite which can easily be run by third-parties.

PR #46, opened 2023-12-11
clean: remove unused hello serviceExport defs

Xunzhuo:
helloServiceExport is not used in anywhere

### DIFF
--- a/conformance/mcs_objects.go
+++ b/conformance/mcs_objects.go
@@ -15,14 +15,3 @@ limitations under the License.
 */
 
 package conformance
-
-import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
-)
-
-var helloServiceExport = v1alpha1.ServiceExport{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "hello",
-	},
-}


mikemorris:
I think this should probably be used at https://github.com/kubernetes-sigs/mcs-api/blob/2833f83cfbdc46d80e0e18111ec977c4ef16e51d/conformance/connectivity_test.go#L92-L93 instead of creating a new ServiceExport inline there.

tpantelis:
> I think this should probably be used at
> 
> https://github.com/kubernetes-sigs/mcs-api/blob/2833f83cfbdc46d80e0e18111ec977c4ef16e51d/conformance/connectivity_test.go#L92-L93
> 
> instead of creating a new ServiceExport inline there.

I think statically defining the resource objects is fine for the two simple tests we have now. But as we expand the tests, we'll likely want to create more than one service and headless services and maybe alter the service spec. So I think defining them globally/statically becomes less useful and maybe cumbersome. I think it's more useful to have functions to create them, eg `createServiceExport(name string)` (which I've added locally in a pending commit).

conformance/mcs_objects.go
```diff
@@ -15,14 +15,3 @@ limitations under the License.
...
-import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
-)
-
-var helloServiceExport = v1alpha1.ServiceExport{
-	ObjectMeta: metav1.ObjectMeta{
-		Name: "hello",
-	},
-}
```
tpantelis, line 28: This would leave an empty file so delete it.

tpantelis:
This PR can be closed in favor of https://github.com/kubernetes-sigs/mcs-api/pull/60 which is a duplicate but it is newer and active.

skitt:
Thanks @Xunzhuo for your contribution; I’ve merged #60 which is more comprehensive.

/close

PR #44, opened 2023-12-11
chore: go mod tidy

Xunzhuo:
chore: go mod tidy

it made failures in make demo:

```
o run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile=./hack/boilerplate.go.txt paths="./..."
../../go/pkg/mod/k8s.io/client-go@v0.25.0/plugin/pkg/client/auth/exec/exec.go:36:2: missing go.sum entry for module providing package golang.org/x/term (imported by k8s.io/client-go/plugin/pkg/client/auth/exec); to add:
        go get k8s.io/client-go/plugin/pkg/client/auth/exec@v0.25.0
Error: not all generators ran successfully
run `controller-gen object:headerFile=./hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=./hack/boilerplate.go.txt paths=./... -h` for usage
exit status 1
make: *** [generate] Error 1
```

PR #47, opened 2023-12-11
fix: code gen path and boilerplate

Xunzhuo:
This should be merged after https://github.com/kubernetes-sigs/mcs-api/pull/44

Currently `make generate` will generate codes in wrong path and boilerplate is not in current year.


Xunzhuo:
/close

Will reopen one, which does not break current path.

PR #49, opened 2024-05-16
update go, deps, and codegen

nojnhuh:
This PR updates Go to 1.22 and k8s.io dependencies to v1.30, handling breaking changes as necessary. 

My original goal here was to get the existing e2e test to run so we can start working on the next one (#16), so there might still be some lingering breakage outside the e2e flow. Hopefully this is at least a step in the right direction.


e2etest/connectivity_test.go
```diff
@@ -102,7 +102,7 @@ var (
 					Containers: []v1.Container{
 						{
 							Name:  "hello-tcp",
-							Image: "alpine/socat",
+							Image: "alpine/socat:1.7.4.4",
```
nojnhuh, line 105: It seems there's an upstream bug in socat causing this to fail: https://github.com/termux/termux-packages/issues/18645

This is the next most recent version, which doesn't seem to have the bug.

jackfrancis: should we use netcat instead? are the modern features in socat vs netcat actually being used in these connectivity tests?

nojnhuh: I don't have a strong opinion, but I was having a bit of trouble wiring up netcat to behave exactly like the UDP socat command. Even with `-k`, `nc -lkup 42 -e echo $MY_POD_ID` was exiting after the first request but the socat command stays alive. Using the same `nc` command without `-u` seems to work fine for TCP.

jackfrancis (approved):
lgtm

AFAICT we should be able to rapidly iterate here on the way towards an eventual v1beta1 release

I'd be delighted to use this PR as a forcing function to accelerate movement towards an sufficent set of E2E outcomes

skitt (approved):

PR #50, opened 2024-05-31
configure coredns with multicluster for e2e

nojnhuh:
This PR extends the existing connectivity test to also verify that the service is reachable via its `*.clusterset.local` domain in addition to its IP. Most of these changes involve setting up CoreDNS with the multicluster plugin: https://github.com/coredns/multicluster


lauralorenz:
AWESOME thank you!

/lgtm
/approve

PR #51, opened 2024-06-04
remove discoveryv1beta1 from e2e tests

nojnhuh:
Non-ancient versions of Kubernetes (1.22-ish? and later) do not serve `v1beta1` of the discovery APIs which include EndpointSlices. This PR removes logic dealing with that API version, which was only present in the connectivity e2e test.

PR #53, opened 2024-06-25
implement test 2: local service not impacted

### LINKED ISSUE #16: Create Test 2 / "local service not impacted" e2e test

nojnhuh:
This PR adds tests to ensure unexported services still behave as expected in the presence of exported ones as described in #16.

Closes #16

### DIFF
--- a/e2etest/connectivity_test.go
+++ b/e2etest/connectivity_test.go
@@ -26,11 +26,7 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
-	discoveryv1 "k8s.io/api/discovery/v1"
-	discoveryv1beta1 "k8s.io/api/discovery/v1beta1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
@@ -56,11 +52,6 @@ var (
 					Protocol: v1.ProtocolUDP,
 				},
 			},
-		},
-	}
-	helloServiceExport = v1alpha1.ServiceExport{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "hello",
 		},
 	}
 	helloServiceImport = v1alpha1.ServiceImport{
@@ -160,6 +151,15 @@ var _ = Describe("Connectivity", func() {
 		serviceImport *v1alpha1.ServiceImport
 		pods          *v1.PodList
 		reqPod        *v1.Pod
+
+		checkPodReachable = func(command []string) {
+			Eventually(func(g Gomega) {
+				stdout, _, err := execCmd(cluster1.k8s, restcfg1, reqPod.Name, reqPod.Namespace, command)
+				g.Expect(err).ToNot(HaveOccurred())
+				ip := strings.TrimSpace(string(stdout))
+				g.Expect(ip).To(Equal(pods.Items[0].Status.PodIP))
+			}, "5m").MustPassRepeatedly(50).Should(Succeed())
+		}
 	)
 	BeforeEach(func() {
 		namespace = fmt.Sprintf("mcse2e-conformance-%v", rand.Uint32())
@@ -179,9 +179,6 @@ var _ = Describe("Connectivity", func() {
 		Expect(err).ToNot(HaveOccurred())
 		svc := helloService
 		_, err = cluster2.k8s.CoreV1().Services(namespace).Create(ctx, &svc, metav1.CreateOptions{})
-		Expect(err).ToNot(HaveOccurred())
-		exp := helloServiceExport
-		_, err = cluster2.mcs.MulticlusterV1alpha1().ServiceExports(namespace).Create(ctx, &exp, metav1.CreateOptions{})
 		Expect(err).ToNot(HaveOccurred())
 		imp := helloServiceImport
 		_, err = cluster1.mcs.MulticlusterV1alpha1().ServiceImports(namespace).Create(ctx, &imp, metav1.CreateOptions{})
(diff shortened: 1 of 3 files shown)


jackfrancis:
@skitt could you kindly give this a review?

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	var (
+		namespace string
+
+		ctx    = context.Background()
+		reqPod *v1.Pod
+	)
+
+	BeforeEach(func() {
```
skitt, line 84: This is fine as-is for now, but one of the design goals we (Paul Morie, @lauralorenz and I) had for the compliance tests was to come up with at least a library of functions to set up and tear down test scenarios. This would be a good start once extracted into a function (or several).

Ultimately we could have a DSL for test scenarios but I think it’s best to evolve that from what actually ends up being useful :wink:.

(Just noting for future reference, please don’t change anything as a result of this comment.)

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+			slices, err = cluster2.k8s.DiscoveryV1().EndpointSlices(namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: labels.Set{discoveryv1.LabelServiceName: helloService.Name}.AsSelector().String(),
+			})
+			Expect(err).ToNot(HaveOccurred())
+			for _, s := range slices.Items {
+				eps += len(s.Endpoints)
+			}
+			return eps
+		}, 30).Should(Equal(1))
+		importedSlice := slices.Items[0]
```
skitt, line 155: It might be worth commenting that this is safe because we know there’s at least one slice (thanks to `Should(Equal(1))` just above).

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			return updatedSlice.Labels[discoveryv1.LabelServiceName]
+		}).ShouldNot(BeEmpty())
+		reqPod, err = cluster1.k8s.CoreV1().Pods(namespace).Get(ctx, requestPod.Name, metav1.GetOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		By("Created all in " + namespace)
+	})
+	AfterEach(func() {
+		if *noTearDown {
+			By(fmt.Sprintf("Skipping tearndown. Test namespace %q", namespace))
```
skitt, line 180:
```suggestion
			By(fmt.Sprintf("Skipping teardown. Test namespace %q", namespace))
```

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+			By(fmt.Sprintf("Cluster 1: kubectl --kubeconfig %q -n %q", *kubeconfig1, namespace))
+			By(fmt.Sprintf("Cluster 2: kubectl --kubeconfig %q -n %q", *kubeconfig2, namespace))
+			return
+		}
+		Expect(cluster1.k8s.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{}))
+		Expect(cluster2.k8s.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{}))
```
skitt, line 186:
```suggestion
		Expect(cluster1.k8s.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{})).To(Succeed())
		Expect(cluster2.k8s.CoreV1().Namespaces().Delete(ctx, namespace, metav1.DeleteOptions{})).To(Succeed())
```

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+	})
+	Specify("DNS resolves as expected", func() {
+		By("verifying a local, unexported service is reachable via cluster.local")
+		successfulChecks := 0
+		command := []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.cluster.local 42", helloService.Name, namespace)}
+		for i := 0; i <= 59; i++ {
+			stout, _, err := execCmd(cluster1.k8s, restcfg1, reqPod.Name, reqPod.Namespace, command)
+			Expect(err).ToNot(HaveOccurred())
+			clusterID := strings.TrimSpace(string(stout))
+			if clusterID == strings.TrimSpace("cluster1") {
```
skitt, line 196:
```suggestion
			if clusterID == "cluster1" {
```

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+			}
+		}
+		Eventually(func() int {
+			return successfulChecks
+		}).Should(BeNumerically(">=", 50))
+
+		By("verifying a remote, exported service is reachable via clusterset.local")
+		successfulChecks = 0
+		command = []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.clusterset.local 42", helloServiceImport.Name, namespace)}
+		for i := 0; i <= 59; i++ {
```
skitt, line 207: This feels like it could be more idiomatically written as a combination of `Eventually` and `MustPassRepeatedly`: first the check must succeed (eventually), then it must pass repeatedly.

The same applies to the similar construct above. The core check should be extracted to a separate function.

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+		command := []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.cluster.local 42", helloService.Name, namespace)}
+		for i := 0; i <= 59; i++ {
+			stout, _, err := execCmd(cluster1.k8s, restcfg1, reqPod.Name, reqPod.Namespace, command)
+			Expect(err).ToNot(HaveOccurred())
+			clusterID := strings.TrimSpace(string(stout))
+			if clusterID == strings.TrimSpace("cluster1") {
+				successfulChecks++
+			}
+		}
+		Eventually(func() int {
```
skitt, line 200: The function’s result can’t change between runs, so `Eventually` isn’t useful.

If the goal is to succeed overall if the connectivity test succeeds 50 times, this should be rewritten as eventually(must succeed repeatedly(…)).

e2etest/localserviceimpact_test.go
```diff
@@ -0,0 +1,280 @@
...
+			eps := 0
+			slices, err = cluster1.k8s.DiscoveryV1().EndpointSlices(namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: labels.Set{discoveryv1.LabelServiceName: helloService.Name}.AsSelector().String(),
+			})
+			Expect(err).ToNot(HaveOccurred())
+			for _, s := range slices.Items {
+				eps += len(s.Endpoints)
+			}
+			return eps
```
skitt, line 232: We’ve seen this before, it should be a function.

skitt (changes requested):
Thanks for this, I have a few requests…

e2etest/localserviceimpact_test.go
nojnhuh: Related to your other comment, I did extract the "export a service from this cluster to that cluster" part of this into its own function.

e2etest/connectivity_test.go
```diff
@@ -205,146 +202,45 @@ var _ = Describe("Connectivity", func() {
 			LabelSelector: metav1.FormatLabelSelector(helloDeployment.Spec.Selector),
 		})
 		Expect(err).ToNot(HaveOccurred())
-		var slicesv1 *discoveryv1.EndpointSliceList
-		var slicesv1b1 *discoveryv1beta1.EndpointSliceList
```
nojnhuh, line 209: These changes now also fulfill #51, so that PR can be closed if this happens to merge first.

skitt:
Thanks! Do please squash these, without even waiting for other reviews — I think the overall result is now easier to review than the individual commits. While you’re at it, if you don’t mind, there are a few `stout` typos instead of `stdout` :wink:.

skitt (approved):

PR #54, opened 2024-07-03
Allow codegen verification

skitt:
Current codegen no longer supports --verify-only; to verify, this generates the code in a temporary directory, and diffs the current code with the newly-generated code. The script exits with the status code of the last command, so it will fail if diff finds any difference.

To correctly account for files which are no longer generated, actual code generation starts by deleting the output directory.


nojnhuh (approved):
/lgtm

PR #57, opened 2024-08-10
Add conformance_suite_test.go generated by ginkgo bootstrap

tpantelis:
...to allow the test suite to be executed via the `ginkgo` CLI in addition to the public entry point in _conformance_suite.go_.

### DIFF
--- a/conformance/conformance_suite_test.go
+++ b/conformance/conformance_suite_test.go
@@ -1,0 +1,13 @@
+package conformance_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


nojnhuh:
See #45

tpantelis:
> See #45

OK instead renaming the file, I added _conformance_suite_test.go_ so the test suite can be run standalone or integrated into an implementation.

tpantelis:
> > See #45
> 
> OK instead renaming the file, I added _conformance_suite_test.go_ so the test suite can be run standalone or integrated into an implementation.

@nojnhuh ☝️

nojnhuh (commented):
/lgtm

PR #59, opened 2024-08-13
Address use of deprecated functions in the conformance package

tpantelis:
See individual commits for details.

### DIFF
--- a/conformance/conformance_suite.go
+++ b/conformance/conformance_suite.go
@@ -18,7 +18,6 @@ package conformance
 
 import (
 	"flag"
-	"math/rand"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


skitt:
/approve

In case other reviewers are wondering about the removal of seed setup, the use of rand in the code doesn’t affect test outcomes, so it doesn’t matter that we can’t fix the seed using Ginkgo.

tpantelis:
> /approve
> 
> In case other reviewers are wondering about the removal of seed setup, the use of rand in the code doesn’t affect test outcomes, so it doesn’t matter that we can’t fix the seed using Ginkgo.

yeah rand is used to generate a unique namespace for each test spec but seeding it isn't necessary.

nojnhuh (commented):
/lgtm

PR #60, opened 2024-08-14
Remove unused helloServiceExport var in conformance

tpantelis:
...and _mcs_objects.go_ as it's now empty.


tpantelis:
> See the discussion in https://github.com/kubernetes-sigs/mcs-api/pull/46

I commented there.

PR #61, opened 2024-08-19
Adjustments to conformance report

tpantelis:
Changed the "Result" column to "Conformant" with values "Yes", "No", and "Unknown". The latter reflects the case where an unexpected error occurs preventing conformance from being determined. The column value can be augmented with additional text describing the reason, eg for "Unknown" the error details would be included.

The test specs report non-conformance via a description function that is passed to Gomaga assertions that is lazily evaluated by the assertion only if a failure occurs. The function adds a report entry named "non-conformant" with an optional message string that is used by the reporting backend to distinguish between a spec failure due to non-conformance or one due to an unexpected error/failure.

Screenshot 2024-08-19 at 10 41 13 AM

### DIFF
--- a/conformance/report.go
+++ b/conformance/report.go
@@ -18,32 +18,68 @@ package conformance
 
 import (
 	_ "embed"
+	"errors"
+	"fmt"
 	"html/template"
 	"os"
+	"regexp"
+	"slices"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/matchers"
 )
 
 const (
-	OptionalLabel      = "Optional"
-	RequiredLabel      = "Required"
-	SpecRefReportEntry = "spec-ref"
+	OptionalLabel            = "Optional"
+	RequiredLabel            = "Required"
+	SpecRefReportEntry       = "spec-ref"
+	NonConformantReportEntry = "non-conformant"
 )
+
+var reportingLabels = []string{
+	RequiredLabel,
+	OptionalLabel,
+}
 
 //go:embed report_template.gohtml
 var reportHTML string
 
 type testInfo struct {
-	Desc string
-	Ref  string
-	Pass bool
+	Desc       string
+	Ref        string
+	Failed     bool
+	Conformant bool
+	Message    string
 }
 
 type testGrouping struct {
 	Name  string
 	Tests []testInfo
+}
+
+var errorRegEx *regexp.Regexp
+
+func init() {
+	dummyErr := errors.New("dummy")
+
+	// Matches gomega output for an error rendered by either the HaveOccurred or Success matchers, eg
+	//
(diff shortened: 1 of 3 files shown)


conformance/report.go
```diff
@@ -55,33 +92,50 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
...
-				Desc: specReport.FullText(),
-				Ref:  ref,
-				Pass: specReport.State == types.SpecStatePassed,
-			})
+			// If the spec failed (ie didn't pass) not due to non-conformance then we assume it encountered
+			// an unexpected error preventing conformance from being determined, and thus we'll report the
+			// conformance status as unknown.
+			if specReport.State != types.SpecStatePassed && info.Conformant {
+				info.Failed = true
+				info.Message = parseFailureMessage(specReport.FailureMessage())
```
jackfrancis, line 125: I wonder if we could generalize `parseFailureMessage` to be `parseSpecReportMessage` (or something), to populate `info.Message` in either the non-conformant or failure case. That way we can drop this:

```
if info.Message != "" {
    info.Message = " - " + info.Message
}
```

and instead put that concatenation into the new, generalized func instead

tpantelis: Non-conformance is determined above via the report entry separate from the Failure message. Re-structuring to a more general `parseSpecReportMessage` would involve more than just populating `info.Message`. Maybe I'm missing your intention - perhaps you could provide a code snippet?

conformance/report.go
```diff
@@ -17,35 +17,72 @@ limitations under the License.
...
 	"html/template"
 	"os"
+	"regexp"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/matchers"
+	"k8s.io/utils/strings/slices"
```
skitt, line 33: Since the PR only uses `slices.Contains()`, this could be `"slices"` instead.

conformance/report.go
```diff
@@ -99,3 +155,44 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
...
+	return firstLine(s)
+}
+
+func firstLine(s string) string {
+	sc := bufio.NewScanner(strings.NewReader(strings.TrimSpace(s)))
+	if sc.Scan() {
+		return strings.TrimSpace(sc.Text())
+	}
+
+	return ""
```
skitt, line 187: Since the strings parsed here are built from string constants (from Ginkgo and elsewhere), this should be safe:

```suggestion
	first, _, _ := strings.Cut(s, "\n")
	return strings.TrimSpace(first)
```

(See https://golang.org/ref/spec#String_literals for context.)

tpantelis: OK - I was trying to be utmost safe.

conformance/report.go
tpantelis: @jackfrancis Does my response ☝️ satisfy your comment?

PR #73, opened 2024-09-16
Rename connectivity_test.go -> connectivity.go

tpantelis:
...so the file is accessible when the conformance module is imported, otherwise the test specs aren't discovered by Ginkgo.


lauralorenz:
/lgtm
though it seems to want a rebase

skitt:
/approve

pending rebase (which will clear lgtms).

PR #76, opened 2024-09-19
Add DNS label to conformance connectivity test

tpantelis:
This test is optional so the specific label will allow users to skip the test rather then have it always fail if not implemented.


MrFreezeex (approved):
/lgtm

MrFreezeex (approved):
/lgtm

Issue #71, opened 2024-09-03
Conformance test: service ports


skitt:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#service-port

Check that Services exported on different ports have the union of service ports exposed.

MrFreezeex:
As I had to implement the port merging recently there is two edge cases that are not very explicit in the KEP that could be checked here:
- Two service ports have the same name in different clusters, something like that:
cluster A:
```
- name: foo
  port: 80
```
cluster B:
```
- name: foo
  port: 8000
```

- Two similar services with different names
cluster A:
```
- name: foo
  port: 80
```
cluster B:
```
- name: bar
  port: 80
```

PR #78, opened 2024-09-30
Add conformance tests for service ports

tpantelis:
When a service is exported from two clusters, ensure the clusterset service exposes the union of service ports declared on its constituent services. Also if port properties don't match, ensure the conflict resolution policy is properly applied.

Creation of K8s resources was refactored to functions rather than defined in global vars to allow the resources to be customized when created on multiple clusters.

Fixes https://github.com/kubernetes-sigs/mcs-api/issues/71

### DIFF
--- a/conformance/service_import.go
+++ b/conformance/service_import.go
@@ -19,17 +19,19 @@ package conformance
 import (
 	"fmt"
 	"net"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
 var _ = Describe("", func() {
 	t := newTestDriver()
 
-	BeforeEach(func() {
-		t.createServiceExport()
+	JustBeforeEach(func() {
+		t.createServiceExport(&clients[0])
 	})
 
 	Specify("Exporting a ClusterIP service should create a ServiceImport of type ClusterSetIP in the service's namespace in each cluster",
@@ -37,7 +39,7 @@ var _ = Describe("", func() {
 			AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services")
 
 			for i := range clients {
-				serviceImport := t.awaitServiceImport(&clients[i], helloService.Name, nil)
+				serviceImport := t.awaitServiceImport(&clients[i], helloServiceName, nil)
 				Expect(serviceImport).NotTo(BeNil(), reportNonConformant(fmt.Sprintf("ServiceImport was not found on cluster %d", i+1)))
 
 				Expect(serviceImport.Spec.Type).To(Equal(v1alpha1.ClusterSetIP), reportNonConformant(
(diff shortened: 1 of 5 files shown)


conformance/service_import.go
```diff
@@ -83,4 +85,71 @@ var _ = Describe("", func() {
...
+				}
+			})
+
+			Specify("should expose the union of the constituent service ports", Label(RequiredLabel), func() {
+				AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port")
+
+				serviceImport := t.awaitServiceImport(&clients[0], t.helloService.Name, func(serviceImport *v1alpha1.ServiceImport) bool {
+					return len(serviceImport.Spec.Ports) == 3
+				})
+				Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
```
skitt, line 125:
```suggestion
				Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found or did not have the expected number of ports")
```

conformance/service_import.go
```diff
@@ -83,4 +85,71 @@ var _ = Describe("", func() {
...
+				Label(RequiredLabel), func() {
+					AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port")
+
+					t.awaitServiceExportCondition(&clients[0], v1alpha1.ServiceExportConflict)
+					t.awaitServiceExportCondition(&clients[1], v1alpha1.ServiceExportConflict)
+
+					serviceImport := t.awaitServiceImport(&clients[0], t.helloService.Name, func(serviceImport *v1alpha1.ServiceImport) bool {
+						return len(serviceImport.Spec.Ports) == len(t.helloService.Spec.Ports)
+					})
+					Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
```
skitt, line 148: Same comment as above.

tpantelis:
> Looks good although I think the conformance port tests would need a bit more edge case test on for conflicting port but this can very well be in another PR (I can do it too if you want) /lgtm /ok-to-test

yeah I saw your comments on the related issue. They're good points but I'm not sure we need to get that granular with these tests. Perhaps it's sufficient to just handle a simple conflict scenario to verify the conflict policy is applied correctly and leave more detailed and/or edge cases to the testing on the implementation's side.

MrFreezeex:
> yeah I saw your comments on the related issue. They're good points but I'm not sure we need to get that granular with these tests. Perhaps it's sufficient to just handle a simple conflict scenario to verify the conflict policy is applied correctly and leave more detailed and/or edge cases to the testing on the implementation's side.

Well the thing is that if you are not careful when you do the union of the ports on the implementation you are likely to construct a ServiceImport with invalid ports (meaning that the API will reject a similar Service with those ports). In a state where this is not handled you would probably only see this in error logs from your controller.

So I do think that it's a good thing to have those in the conformance test in some way as it would ensure that those case are handled properly by the implementers.

conformance/service_import.go
tpantelis: `awaitServiceImport` only returns nil if the `ServiceImport` is not found. If the verification function isn't satisfied then `awaitServiceImport` times out but still succeeds and returns the `ServiceImport`, enabling more granular and exact assertions by the caller with richer descriptions and avoiding vaguer description as suggested above.

tpantelis:
> > yeah I saw your comments on the related issue. They're good points but I'm not sure we need to get that granular with these tests. Perhaps it's sufficient to just handle a simple conflict scenario to verify the conflict policy is applied correctly and leave more detailed and/or edge cases to the testing on the implementation's side.
> 
...
> So I do think that it's a good thing to have those in the conformance test in some way as it would ensure that those case are handled properly by the implementers.

yeah it's a fair point and discussion to have but that seems to be more of taking on the role of exhaustive E2E tests. I think we should keep the focus on MCS API/spec compliance/conformance. Validity of service ports and edge cases/caveats thereof is more core K8s and certainly can be mentioned in the MCS spec or at least linked to in core K8s docs.

MrFreezeex:
> think we should keep the focus on MCS API/spec compliance/conformance. Validity of service ports and edge cases/caveats thereof is more core K8s and certainly can be mentioned in the MCS spec or at least linked to in core K8s docs.

Well to me getting the controller to produce an invalid ServiceImport is a conformance problem. I don't think that the conformance tests should only tests the happy path/easy cases. 

And even past that if there are potential caveat that affect most implementations why not add it there too? It helps level up all the implementations by providing them cases that they didn't necessarily think...

tpantelis:
> > think we should keep the focus on MCS API/spec compliance/conformance. Validity of service ports and edge cases/caveats thereof is more core K8s and certainly can be mentioned in the MCS spec or at least linked to in core K8s docs.
> 
> Well to me getting the controller to produce an invalid ServiceImport is a conformance problem. I don't think that the conformance tests should only tests the happy path/easy cases.
> 
> And even past that if there are potential caveat that affect most implementations why not add it there too? It helps level up all the implementations by providing them cases that they didn't necessarily think of...

Good points, although should we assume that what constitutes a valid clusterset `ServiceImport` necessarily matches what constitutes a valid cluster `Service`?... I mean is the union of the ports necessarily valid? Say cluster1 exports service port TCP/100 but cluster2 doesn't - what should happen if a client is resolved to cluster2's service IP and tries to access port TCP/100? I guess this depends on the implementation (_Submariner_ internally exposes the intersection of the ports).

Anyway, we can certainly add more tests to cover cases as you suggest to conform to the spec. But, right now, the spec only talks about conflicting port names, hence why I implemented that scenario in the tests. Perhaps we should first update the spec as to exactly what scenarios constitute a `ServiceImport` port conflict and then consider adding additional conformance tests.

MrFreezeex:
> I mean is the union of the ports necessarily valid? Say cluster1 exports service port TCP/100 but cluster2 doesn't - what should happen if a client is resolved to cluster2's service IP and tries to access port TCP/100? I guess this depends on the implementation (_Submariner_ internally exposes the intersection of the ports).

Well the KEP state explicitly that it should be the union so I believe Submariner won't be conformant in that sense to the current KEP :/. 

> Anyway, we can certainly add more tests to cover cases as you suggest to conform to the spec. But, right now, the spec only talks about conflicting port names, hence why I implemented that scenario in the tests. Perhaps we should first update the spec as to exactly what scenarios constitute a ServiceImport port conflict and then consider adding additional conformance tests.

Yes it's true that the KEP could be updated although to me this is almost common sense that the ports in ServiceImport should respect the ports rules of regular Services. If we want to keep a test very very generic it could not even test for conflict directly but just throws a few ports at it and tests that the resulting ports are valid (aka no duplicated name and no duplicated ports with the same port number and proto).

tpantelis:
> > I mean is the union of the ports necessarily valid? Say cluster1 exports service port TCP/100 but cluster2 doesn't - what should happen if a client is resolved to cluster2's service IP and tries to access port TCP/100? I guess this depends on the implementation (_Submariner_ internally exposes the intersection of the ports).
> 
> Well the KEP state explicitly that it should be the union so I believe Submariner won't be conformant in that sense to the current KEP :/.
> 

Well, Submariner does expose the union as per the spec wrt the `ServiceImport` CR but, under the hood, it's the intersection that's used b/c it wouldn't be valid otherwise (unless the DNS query specifies the cluster name). I don't know how other implementations handle this. I brought this up at the sig meeting a while back and, FWIR, it was acknowledged by the original authors that the union could be problematic but that's what was decided at that time. 

> Yes it's true that the KEP could be updated although to me this is almost common sense that the ports in ServiceImport should respect the ports rules of regular Services. If we want to keep a test very very generic it could not even test for conflict directly but just throws a few ports at it and tests that the resulting ports are valid (aka no duplicated name and no duplicated ports with the same port number and proto).

Perhaps it is common sense but I was just going by the current spec. However it may not be common knowledge what the exact ports rules are for regular services. It sounds like you performed actual testing to find that out. I still think it would be useful to update the spec to make it clear exactly what constitutes a valid `ServiceImport` port. Maybe the original authors were aware of the exact service rules but, for some reason, decided to just enumerate port name equality for `ServiceImport`.

tpantelis:
The `pull-mcs-api-e2e` job is intermittently failing

MrFreezeex (approved):
/lgtm

Issue #63, opened 2024-08-20
Research and update MCS API with the current situation for dual stack


lauralorenz:
The `ServiceImport.Spec.IPs` field ([ref](https://github.com/kubernetes-sigs/mcs-api/blob/master/pkg/apis/v1alpha1/serviceimport.go#L56-L58)) was set as an array to support dual stack networkign which, at the time, was an in flight project by SIG-Network. From a cursory glance at the `Service.Spec` [API reference docs](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) it looks like we may be out of date with how that shaped up. For this issue, please research the current API formulation from SIG-Network to store dual stack IPs and update the MCS API `ServiceImport.Spec.IPs` field to match.

PR #80, opened 2024-10-04
Update service controller to add all the IPs from derived service

MrFreezeex:
Update the Service controller to update the ServiceImport with all the IPs instead of only the first IP.

A bit related to #63 might not be the complete answers though

### DIFF
--- a/pkg/controllers/service.go
+++ b/pkg/controllers/service.go
@@ -68,7 +68,7 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
-	svcImport.Spec.IPs = []string{service.Spec.ClusterIP}
+	svcImport.Spec.IPs = service.Spec.ClusterIPs
 	if err := r.Client.Update(ctx, &svcImport); err != nil {
 		return ctrl.Result{}, err
 	}


pkg/controllers/service.go
```diff
@@ -68,7 +68,7 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
-	svcImport.Spec.IPs = []string{service.Spec.ClusterIP}
+	svcImport.Spec.IPs = service.Spec.ClusterIPs
```
tpantelis, line 71: This may be problematic since the CRD limits the array to [one item](https://github.com/kubernetes-sigs/mcs-api/blob/eff1ba8c3ab2d50a7faa0a02d18dc6ef9f58a21d/pkg/apis/v1alpha1/serviceimport.go#L56).

MrFreezeex: Ah yes indeed good point thanks! The alpha2 version on #52 does make this to two items max so I am thinking that we could just wait for this version somehow?

cc @mikemorris

mikemorris:
```suggestion
	svcImport.Spec.IPs = service.Spec.ClusterIPs[:2]
```
This field on ServiceImport is intended for parity with Service `ClusterIPs` (as noted in #63), which is [restricted to two items](https://github.com/kubernetes/kubernetes/blob/a499facee693a1a83daadb82d88f7b51d324ffc5/pkg/apis/core/types.go#L4688-L4696) but we still might want an extra safeguard here?

I do think this is blocked by #85 updating the max length to 2 as @tpantelis noted, but once that is in this LGTM.

mikemorris (approved):
/approve
/hold

Pending comments in https://github.com/kubernetes-sigs/mcs-api/pull/80/files#r1878240881, #85 should merge first

MrFreezeex:
> Sorry I've been behind following up on your work! This feels okay to me, may need to regenerate the CRD YAML in this PR now though.

No worries, actually I have already regenerated the yaml so it should be fine on that aspect :+1:

pkg/controllers/service.go
MrFreezeex: FTR you can't do this `service.Spec.ClusterIPs[:2]` if the the slice capacity is smaller than 2, I tweaked it to make this work by checking the len first though!

PR #81, opened 2024-10-24
Extract cluster name from kube config for conformance reporting

tpantelis:
...in lieu of reporting the internal cluster index which isn't very useful to the end user.


MrFreezeex (approved):
Looks very nice indeed, thanks! :D 
/lgtm

Issue #66, opened 2024-09-03
Conformance test: importing services


skitt:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#importing-services

Using DNS (since accessing exported services relies on DNS), check that:

- [x] Exporting a ServiceExport results in the creation of a corresponding ServiceImport in each importing cluster hosting the corresponding namespace
- [ ] Importing clusters without the corresponding namespace don’t get a ServiceImport (the namespace may be created automatically, so the test can’t assume that because it hasn’t created the namespace, the namespace doesn’t exist)
- [x] ServiceImports only exist as long as at least one corresponding source ServiceExport exists
- [x] EndpointSlices are labeled as follows: “multicluster.kubernetes.io/service-name” pointing to the ServiceImport, “multicluster.kubernetes.io/source-cluster” pointing to the source cluster

tpantelis:
> * [ ]  Exporting a ServiceExport results in the creation of a corresponding ServiceImport in each importing cluster hosting the corresponding namespace

This was done as part of https://github.com/kubernetes-sigs/mcs-api/pull/64.

> * [ ]  Importing clusters without the corresponding namespace don’t get a ServiceImport (the namespace may be created automatically, so the test can’t assume that because it hasn’t created the namespace, the namespace doesn’t exist)

This test would have to be optional and should first wait a reasonable of time for the namespace to be created and, if not, then skip the test. It can also have a specific label (eg "Namespace") so users can skip it if they know their implementation doesn't support auto-creating namespaces.

> * [ ]  EndpointSlices are labeled as follows: “multicluster.kubernetes.io/service-name” pointing to the ServiceImport, “multicluster.kubernetes.io/source-cluster” pointing to the source cluster

This should be covered by https://github.com/kubernetes-sigs/mcs-api/issues/70.

skitt:
> > * [ ]   Importing clusters without the corresponding namespace don’t get a ServiceImport (the namespace may be created automatically, so the test can’t assume that because it hasn’t created the namespace, the namespace doesn’t exist)
> 
> This test would have to be optional and should first wait a reasonable of time for the namespace to be created and, if not, then skip the test. It can also have a specific label (eg "Namespace") so users can skip it if they know their implementation doesn't support auto-creating namespaces.

[...]

PR #82, opened 2024-10-31
Add conformance tests for service unexport

tpantelis:
Ensure corresponding ServiceImport is deleted.

Related to https://github.com/kubernetes-sigs/mcs-api/issues/66

### DIFF
--- a/conformance/service_import.go
+++ b/conformance/service_import.go
@@ -34,7 +34,8 @@ var _ = Describe("", func() {
 		t.createServiceExport(&clients[0])
 	})
 
-	Specify("Exporting a ClusterIP service should create a ServiceImport of type ClusterSetIP in the service's namespace in each cluster",
+	Specify("Exporting a ClusterIP service should create a ServiceImport of type ClusterSetIP in the service's namespace in each cluster. "+
+		"Unexporting should delete the ServiceImport",
 		Label(RequiredLabel), func() {
 			AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services")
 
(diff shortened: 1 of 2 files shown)


conformance/service_import.go
```diff
@@ -87,6 +97,44 @@ var _ = Describe("", func() {
 				reportNonConformant(fmt.Sprintf("The value %q is not a valid IP", serviceImport.Spec.IPs[0])))
 		})
 
+	Specify("A ServiceImport should only exist as long as there's at least one exporting cluster",
```
MrFreezeex, line 100: Maybe we could better emphasize that here we are testing (could be via a code comment not necessarily the `Specify` I guess) the deletion with two clusters exporting a service compared to the previous test?

conformance/service_import.go
```diff
@@ -87,6 +97,44 @@ var _ = Describe("", func() {
...
+			Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
+
+			By(fmt.Sprintf("Exporting the service on the second cluster %q", clients[1].name))
+
+			t.deployHelloService(&clients[1], newHelloService())
+			t.createServiceExport(&clients[1])
+
+			// Sanity check and to also wait a bit for the second cluster to export. There's no deterministic way to tell if/when the
+			// second cluster has finished exporting other than utilizing different service ports in each cluster but that's already
+			// done in another test case.
```
MrFreezeex, line 118: `other than utilizing different service ports in each cluster but that's already done in another test case.` 
Is that a future improvement that could be done here too? AFAICT this is not checked in the code below, if that's the case maybe this could be marked as FIXME/TODO or something like that?

MrFreezeex (commented):
Two small comments but overall this looks good thanks :pray:!

conformance/service_import.go
tpantelis: I could remove the unexport part from the previous test but I added it there b/c that test doesn't require two clusters and I think it's good that we try to test as much as we can if only one cluster is specified. The distinction with this test is that it ensures the SI isn't deleted when the first cluster service is unexported.  I think that "_as long as there's at least one exporting cluster_" implies the distinction but, either way, I can add a code comment stating the above.

conformance/service_import.go
tpantelis: I wanted to avoid this test relying on the service port behavior as that is tested on L168 below, otherwise if there's an issue with service port compliance than this test would also fail orthogonally.

  > Is that a future improvement that could be done here too? AFAICT this is not checked in the code below, if that's the case maybe this could be marked as FIXME/TODO or something like that?

I assume you're referring to the L168 test that doesn't unexport and verify the ports are adjusted correctly. I could add that but I also was wary of that failing orthogonally if the SI was unexpectedly deleted.

conformance/service_import.go
MrFreezeex: I think both of your suggestions are fine! My main concern was that if you read only the specify you could think that one of the test may be redundant so just slightly clarifying this would be awesome!

conformance/service_import.go
MrFreezeex: Ah ok understood! Maybe it would be nice to clarify this in the comment as the part `but that's already done in another test case.` is not 100% clear to me as to why this was not done without your additional explanation?


Maybe something like: `second cluster has finished exporting other than utilizing different service ports in each cluster but the scope of this test is not about testing a valid port merging implementation.`? (Feel free to update this, it's just a proposal)

MrFreezeex (approved):
Thanks again!
/lgtm

skitt:
I just added a missing space.

/lgtm
/approve

Issue #69, opened 2024-09-03
Conformance test: DNS


skitt:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#dns

Check that:

- [ ] Exported services are accessible using service.ns.svc.clusterset.local
- [ ] Headless services are represented by multiple A/AAAA records
- [ ] SRV records are available
- [ ] Headless services’ individual pods are accessible through hostname.clusterid.svc.ns.svc.clusterset.local
- [ ] The cluster.local zone is unchanged, and accessing services there only accesses local services
- [ ] Since .clusterset.local may be customized, the tests should allow it to be overridden

tpantelis:
@skitt These tasks can be checked as complete (I don't have privs):

- Exported services are accessible using service.ns.svc.clusterset.local
- SRV records are available
- The cluster.local zone is unchanged, and accessing services there only accesses local services

skitt:
#109 tracks AAAA records, thanks. I’ll open another issue regarding `.clusterset.local`.

PR #83, opened 2024-11-27
Add DNS conformance tests for ClusterIP services

tpantelis:
- Verify DNS lookup of the .clusterset.local domain returns the clusterset IP
- Verify DNS SRV query
- Verify DNS lookup of the .cluster.local domain returns the local cluster IP

Related to https://github.com/kubernetes-sigs/mcs-api/issues/69

The existing connectivity test was changed to specify "_Connectivity_" label instead of the "_DNS_" label to allow more granular filtering. An implementation may implement DNS lookup of the clusterset IP but not routing across clusters.

### DIFF
--- a/conformance/clusterip_service_dns.go
+++ b/conformance/clusterip_service_dns.go
@@ -1,0 +1,160 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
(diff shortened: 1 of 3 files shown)


tpantelis:
cc @MrFreezeex

MrFreezeex (approved):
Thanks! The code looks good and I confirmed that it works with the Cilium implementation (and thus the multicluster coredns plugin).

/lgtm

tpantelis:
@skitt Can you review when you get a chance?

conformance/clusterip_service_dns.go
```diff
@@ -0,0 +1,160 @@
...
+		})
+
+		Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
+		Expect(serviceImport.Spec.IPs).ToNot(BeEmpty(), "ServiceImport does not contain an IP")
+
+		clusterSetIP := serviceImport.Spec.IPs[0]
+
+		By(fmt.Sprintf("Found ServiceImport with clusterset IP %q", clusterSetIP))
+
+		command := []string{"sh", "-c", fmt.Sprintf("nslookup %s.%s.svc.clusterset.local", t.helloService.Name, t.namespace)}
```
skitt, line 56: Strictly speaking this doesn’t need a shell but that’s a minor issue.

PR #84, opened 2024-12-02
apis: conformance: add labels and annotation export

MrFreezeex:
This adds annotations and labels exports fields and related conformance related to https://github.com/kubernetes/enhancements/pull/4922. This add the fields directly into v1alpha1 as those fields are optionally implemented and additive meaning this change should be retro compatible.

**This shouldn't be merged until the KEP change is also merged**


MrFreezeex:
This is in draft because the associated KEP is not merged yet, but apart from that it should be ready I believe

cc @tpantelis @skitt if you could review this in the meantime that would be awesome :pray:

conformance/service_import.go
```diff
@@ -204,5 +222,50 @@ var _ = Describe("", func() {
...
+	})
+
+	Context("", func() {
+		BeforeEach(func() {
+			helloServiceExport.Spec.ExportedAnnotations = map[string]string{"dummy-annotation": "true"}
+			helloServiceExport.Spec.ExportedLabels = map[string]string{"dummy-label": "true"}
+		})
+
+		Specify("export labels and annotations",
+			Label(OptionalLabel), func() {
```
tpantelis, line 259: Let's define a specific label as well so this test can be skipped if the implementation doesn't support exported labels.

`const ExportedLabelsLabel = "ExportedLabels"`

```suggestion
			Label(OptionalLabel), Label(ExportedLabelsLabel), func() {
```

conformance/service_import.go
```diff
@@ -204,5 +222,50 @@ var _ = Describe("", func() {
...
+				AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#labels-and-annotations")
+
+				serviceImport := t.awaitServiceImport(&clients[0], helloServiceName, func(serviceImport *v1alpha1.ServiceImport) bool {
+					return len(serviceImport.Labels) > 0
+				})
+				Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
+
+				Expect(serviceImport.Annotations).To(Equal(helloServiceExport.Annotations), reportNonConformant(""))
+				Expect(serviceImport.Labels).To(Equal(helloServiceExport.Labels), reportNonConformant(""))
```
tpantelis, line 268: Shouldn't these check the Spec fields?

```suggestion
				Expect(serviceImport.Annotations).To(Equal(helloServiceExport.Spec.ExportedAnnotations), reportNonConformant(""))
				Expect(serviceImport.Labels).To(Equal(helloServiceExport.Spec.ExportedLabels), reportNonConformant(""))
```

conformance/service_import.go
```diff
@@ -139,13 +144,26 @@ var _ = Describe("", func() {
 				reportNonConformant(fmt.Sprintf("The value %q is not a valid IP", serviceImport.Spec.IPs[0])))
 		})
 
+	Specify("Annotations and labels should be empty if not explicitly exported",
```
tpantelis, line 147: Test cases require creating a namespace and other resources so they aren't cheap. As such, I don't think we need a specific test case for this. This would naturally be verified as part of the spec on L258 below.

conformance/service_import.go
```diff
@@ -204,5 +222,50 @@ var _ = Describe("", func() {
...
+		})
+	})
+
+	Context("", func() {
+		BeforeEach(func() {
+			helloServiceExport.Spec.ExportedAnnotations = map[string]string{"dummy-annotation": "true"}
+			helloServiceExport.Spec.ExportedLabels = map[string]string{"dummy-label": "true"}
+		})
+
+		Specify("export labels and annotations",
```
tpantelis, line 258: Let's make this more descriptive, eg "_Labels and annotations specified as exported in the ServiceExport should be propagated to the ServiceImport_".

tpantelis (changes requested):

conformance/service_import.go
MrFreezeex: Woops indeed my bad :man_facepalming:

conformance/service_import.go
MrFreezeex: Makes sense! I removed it

config/crd/multicluster.x-k8s.io_serviceexports.yaml
```diff
@@ -58,6 +58,19 @@ spec:
...
+            spec:
+              description: spec defines the behavior of a ServiceExport.
+              type: object
+              properties:
+                exportedAnnotations:
+                  type: object
+                  additionalProperties:
+                    type: string
+                exportedLabels:
+                  description: ExportedLabels describes the annotations exported. It is optional for implementation.
```
giorio94, line 70: Nit.

```suggestion
                  description: ExportedLabels describes the labels exported. It is optional for implementation.
```

config/crd/multicluster.x-k8s.io_serviceexports.yaml
```diff
@@ -58,6 +58,19 @@ spec:
               type: string
             metadata:
               type: object
+            spec:
+              description: spec defines the behavior of a ServiceExport.
+              type: object
+              properties:
+                exportedAnnotations:
```
giorio94, line 65: Nit. The description is not being rendered correctly.

MrFreezeex:
Undrafted (since the KEP change should be merged soon hopefully) and rebased/fixed the new calls to createServiceExport that was merged in the meantime

MrFreezeex:
/cc @JeremyOT @skitt 

Could you take a look at this PR 🙏 , it's the last piece to unblock the cilium related PR 😅

PR #85, opened 2024-12-03
apis/v1alpha2: move ServiceImport spec fields to status

MrFreezeex:
This PR is meant to supersedes #52, most commits are from this linked PR done by @mikemorris (kudos to him!) and with the few suggestions left applied and making sure everything generate well


MrFreezeex:
While this adds initial support for v1alpha2 in this repo the next step involves:
- Updating the coredns/multicluster plugin to detect which version is installed and selectively watch one api or the other (so that new coredns version are still compatible with v1alpha1)
- Update e2e tests  (that depends on the coredns plugin) and storage version to v1alpha2
- Possibly updating conformance tests to this versions (cc @tpantelis for opinion on this aspect?)
- Figuring out the code/recommendation for implementations for the webhooks

If possible I would prefer to have #84 merged first and then updates this PR with the new fields since we would like to have #84 in Cilium and there is a bit of things left to do to have a v1alpha2 ready (especially the coredns plugin) which means that unfortunately v1alpha2 might not be able to make it in time for the next version of Cilium.

pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -23,6 +23,8 @@ import (
 
 // +genclient
 // +kubebuilder:object:root=true
+// +kubebuilder:storageversion
```
tpantelis, line 26: Shouldn't the new _v1alpha2_ be the stored version?

MrFreezeex: The problem is that the CoreDNS plugin or anything watching mcs api ressources would need to support v1alpha2 explicitly or that we have the webhooks already deployed/figured out. So my plan would be to do this after CoreDNS supports v1alpha2, see this https://github.com/kubernetes-sigs/mcs-api/pull/85#issuecomment-2514593992

config/crd-base/multicluster.x-k8s.io_serviceimports.yaml
tpantelis, line 1: What is this file for?

config/crd/multicluster.x-k8s.io_serviceimports.yaml
```diff
@@ -1,183 +1,351 @@
...
+            enum:
+            - ClusterSetIP
+            - Headless
+            type: string
+        required:
+        - ports
+        - type
+        type: object
+    served: true
+    storage: false
```
tpantelis, line 351: This file does not specify any webhook conversion config so who is the intended consumer?

tpantelis:
> While this adds initial support for v1alpha2 in this repo the next step involves:
> 
> * Possibly updating conformance tests to this versions (cc @tpantelis for opinion on this aspect?)

My thoughts are that the consuming test code can be updated to reference the  _v1alpha2_ Go versions.  At test suite setup, we read the CRD and and check which versions are served and create dynamic `ResourceClients` with the selected GVRs (biasing towards _v1alpha2_). The functions to read/create `ServiceImports`/`ServiceExports` would use the dynamic clients instead of the generated client sets. If _v1alpha2_ is served in the CRD then we can directly convert the `Unstructured` to the _v1alpha2_ Go struct (and vice versa). Otherwise, we first convert  the _v1alpha1_  `Unstructured` to _v1alpha2_ (using the same code as the webhook converter) and then convert  to the _v1alpha2_ Go struct.

config/crd-base/multicluster.x-k8s.io_serviceimports.yaml
MrFreezeex: I am not entirely sure tbh, it was part of the crd generation before 🤷‍♂️

config/crd/multicluster.x-k8s.io_serviceimports.yaml
MrFreezeex: I wanted to figure out webhooks as a later steps when we would be actually storing v1alpha2, but we also said at the last sig meeting that we could provide only the webhook code and not the deployment manifests so I suppose that implementation would have to update that aspect if they want to have a webhook. TBD when we would be actually implementing that and figured out recomendations to integrate it to implementations.

tpantelis: I can work on the webhook code - shouldn't be difficult.

MrFreezeex: Sure that would be great! Note that I would like to submit the CoreDNS PR to v1alpha2 as soon as possible with the hope that it gets merged before the winter holidays by the CoreDNS folks. So I would recommend to not block the v1alpha2 version based on having the webhook or not.

pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -23,6 +23,8 @@ import (
...
+// +kubebuilder:resource:shortName=svcimport
```
skitt, line 27: The short name for service imports is currently scvim (see the CRDs).

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -22,6 +22,8 @@ import (
 
 // +genclient
 // +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +kubebuilder:resource:shortName=svcexport
```
skitt, line 26: The short name for service exports is currently svcex.

skitt:
@MrFreezeex could you rebase this?

config/crd/multicluster.x-k8s.io_serviceimports.yaml
zhiying-lin: @tpantelis not sure if you have started looking at the wehbook or not. If not, i'm willing to help here.

tpantelis: I have recently looked into conversion webhooks for another reason but concluded that, while they are fairly straightforward to implement, they are problematic for deployment due to the required TLS certificate, which makes seamless upgrades challenging. As such, we decided to keep CRD changes backwards compatible and avoid a webhook.

IMO, there has to be a really compelling functional reason to break backwards compatibility requiring a webhook converter. I don't think restructuring the ServiceImport spec fields for purely theoretical/philosophical reasons fits that criteria and it will put unnecessary burden on implementors to deal with.

zhiying-lin: @tpantelis Could you please clarify this "they are problematic for deployment due to the required TLS certificate, which makes seamless upgrades challenging. "? why it's related to TLS. once we have that mutation webhook, api server should return you the version you request no matter what's stored in the etcd.

lauralorenz:
Triage notes: cross reference with the slack thread to make sure that this is the correct format of the PR (about the specific destination of the content change (root vs status) and about what version of API that should actually be under).

lauralorenz:
/hold

Until we confirm with the slack thread that this is the correct PR for people to review

PR #86, opened 2024-12-06
Fail early if the MCS CRDs are not available

skitt:
Explicitly list ServiceExports and ServiceImports, to make sure that all named contexts are reachable and have the necessary CRDs.

Accumulate all errors in setupClients() so that the user knows about errors in all named contexts instead of just the first one.

### DIFF
--- a/conformance/conformance_suite.go
+++ b/conformance/conformance_suite.go
@@ -83,45 +83,62 @@ var _ = BeforeSuite(func() {
 func setupClients() error {
 	splitContexts := strings.Split(contexts, ",")
 	clients = make([]clusterClients, len(splitContexts))
-	for i, context := range splitContexts {
-		overrides := clientcmd.ConfigOverrides{ClusterDefaults: clientcmd.ClusterDefaults}
-		overrides.CurrentContext = context
+	accumulatedErrors := []error{}
 
-		clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &overrides)
+	for i, kubeContext := range splitContexts {
+		err := func() error {
+			overrides := clientcmd.ConfigOverrides{ClusterDefaults: clientcmd.ClusterDefaults}
+			overrides.CurrentContext = kubeContext
 
-		rawConfig, err := clientConfig.RawConfig()
-		if err != nil {
-			return err
-		}
+			clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &overrides)
 
-		name := context
-		if name == "" {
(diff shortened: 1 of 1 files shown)


conformance/conformance_suite.go
```diff
@@ -106,22 +109,42 @@ func setupClients() error {
...
+		if _, err := mcsClient.MulticlusterV1alpha1().ServiceExports("").List(context.TODO(), metav1.ListOptions{}); err != nil {
+			accumulatedErrors = append(accumulatedErrors,
+				fmt.Errorf("error listing ServiceExports on context %s, is the MCS API installed? %w", name, err))
+			continue
+		}
+
+		if _, err := mcsClient.MulticlusterV1alpha1().ServiceImports("").List(context.TODO(), metav1.ListOptions{}); err != nil {
+			accumulatedErrors = append(accumulatedErrors,
+				fmt.Errorf("error listing ServiceImports on context %s, is the MCS API installed? %w", name, err))
+			continue
```
tpantelis, line 137: To support multiple version in the future, we'll likely retrieve the CRD resources but this is fine for now.

conformance/conformance_suite.go
```diff
@@ -106,22 +109,42 @@ func setupClients() error {
...
 
 		clients[i] = clusterClients{name: name, k8s: k8sClient, mcs: mcsClient, rest: restConfig}
 	}
-	return nil
+
+	if len(accumulatedErrors) == 0 {
+		return nil
+	} else {
+		return errors.Join(accumulatedErrors...)
+	}
```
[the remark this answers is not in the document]
skitt: So it does, and I’ve used that feature before…

conformance/conformance_suite.go
```diff
@@ -83,45 +83,62 @@ var _ = BeforeSuite(func() {
...
-		}
+			clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &overrides)
 
-		name := context
-		if name == "" {
-			name = rawConfig.CurrentContext
-		}
+			rawConfig, err := clientConfig.RawConfig()
+			if err != nil {
+				return fmt.Errorf("error setting up a Kubernetes API client on context %s: %w", kubeContext, err)
```
tpantelis, line 97: `errors.Wrap` is a bit simpler. Same below.

```suggestion
				return errors.Wrapf(err, "error setting up a Kubernetes API client on context %q", kubeContext)
```

skitt: I know, but it’s not _that_ much simpler and we already have a different `errors` import…

Issue #74, opened 2024-09-17
pull-mcs-api-verify job reports success but is actually failing


nojnhuh:
The `golint` step of the pull-mcs-api-verify presubmit job seems to be failing but that isn't causing the whole job to fail as I think it should:

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_mcs-api/72/pull-mcs-api-verify/1835855230667853824

```
no required module provides package sigs.k8s.io/mcs-api/cmd/servicecontroller; to add it:
	go get sigs.k8s.io/mcs-api/cmd/servicecontroller
no required module provides package sigs.k8s.io/mcs-api/conformance; to add it:
	go get sigs.k8s.io/mcs-api/conformance
no required module provides package sigs.k8s.io/mcs-api/e2etest; to add it:
	go get sigs.k8s.io/mcs-api/e2etest
no required module provides package sigs.k8s.io/mcs-api/pkg/apis/v1alpha1; to add it:
	go get sigs.k8s.io/mcs-api/pkg/apis/v1alpha1
no required module provides package sigs.k8s.io/mcs-api/pkg/client/clientset/versioned; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/clientset/versioned
no required module provides package sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/fake; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/fake
no required module provides package sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/scheme; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/scheme
no required module provides package sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1
no required module provides package sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1/fake; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/clientset/versioned/typed/apis/v1alpha1/fake
no required module provides package sigs.k8s.io/mcs-api/pkg/client/informers/externalversions; to add it:
	go get sigs.k8s.io/mcs-api/pkg/client/informers/externalversions
[...]

PR #87, opened 2024-12-06
Fix Go linting

skitt:
The switch to the tools module for tool dependency tracking broke verify-golint.sh (golint was run from tools by "go -C tools" but the packages were no longer accessible).

This fixes the above by converting the packages to relative references.

It also ensures that output sent to stderr is not ignored, so that future errors of this type don't go undetected. It also ignores the golint error about OptionalLabel; documenting that isn't very useful.

Other commits fix the linting errors that previously went undetected.

Fixes: #74

### DIFF
--- a/conformance/framework.go
+++ b/conformance/framework.go
@@ -19,88 +19,14 @@ package conformance
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"time"
 
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


conformance/framework.go
```diff
@@ -34,28 +34,30 @@ import (
...
+// AwaitUntil repeats the given operation until the given checkResultFunc returns true, an error, or a
+// timeout is reached; it then verifies that no error occurred.
+func AwaitUntil(opMsg string, doOperation doOperationFunc, checkResult checkResultFunc) interface{} {
 	result, errMsg, err := AwaitResultOrError(opMsg, doOperation, checkResult)
 	Expect(err).NotTo(HaveOccurred(), errMsg)
 
 	return result
 }
 
-func AwaitResultOrError(opMsg string, doOperation DoOperationFunc, checkResult CheckResultFunc) (interface{}, string, error) {
```
tpantelis, line 50: I think we can get rid of these two "Await" functions as they're not used and I don't see a need for them.

Issue #65, opened 2024-09-03
Conformance test: exporting services


skitt:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#exporting-services

Check that:

- [ ] The ServiceExport CRD exists
- [ ] Creating a ServiceExport exports the name-mapped Service (i.e. the service becomes accessible from importing clusters, see the “importing services” test for the DNS entries)
- [ ] If a ServiceExport exists on cluster A, and not on cluster B, and the corresponding service exists on both, remote clusters only access cluster A, never cluster B
- [ ] Name-mapped Services are only exported as long as the corresponding ServiceExport exists

tpantelis:
@skitt These tasks can be checked as complete:

- Name-mapped Services are only exported as long as the corresponding ServiceExport exists
- The ServiceExport CRD exists

There's no explicit test for the latter but it's indirectly verified since creation of `ServiceExport` would fail if the CRD wasn't installed. However we might retrieve the CRD to check the versions if we want to support _v1alpha1_ and _v1alpha2_ in the future.

This task is done for ClusterIP services but not headless yet so we can keep it open tho I'm fine with marking as complete:

- Creating a ServiceExport exports the name-mapped Service (i.e. the service becomes accessible from importing clusters, see https://github.com/kubernetes-sigs/mcs-api/issues/69 for the DNS entries)

skitt:
@tpantelis Regarding the ServiceExport CRD, I was thinking of adding an early check because currently, the conformance suite attempts to run to completion even on a cluster with no MCS controller, which could be confusing. See #86

tpantelis:
These 2 can now be marked as done:

- The ServiceExport CRD exists
- If a ServiceExport exists on cluster A, and not on cluster B, and the corresponding service exists on both, remote clusters only access cluster A, never cluster B

PR #89, opened 2024-12-06
Add connectivity conformance test involving service in two clusters

tpantelis:
...as described in https://github.com/kubernetes-sigs/mcs-api/issues/65:
    
"_If a ServiceExport exists on cluster A, and not on cluster B, and the corresponding service exists on both, remote clusters only access cluster A, never cluster B_"

The first commit improves the error reporting when verifying command exec output. See commit message for details.

### DIFF
--- a/conformance/connectivity.go
+++ b/conformance/connectivity.go
@@ -18,6 +18,7 @@ package conformance
 
 import (
 	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


conformance/connectivity.go
```diff
@@ -69,4 +70,52 @@ var _ = Describe("Connectivity to remote services", func() {
...
+				pods, err := clients[0].k8s.CoreV1().Pods(t.namespace).List(context.TODO(), metav1.ListOptions{
+					LabelSelector: metav1.FormatLabelSelector(newHelloDeployment().Spec.Selector),
+				})
+				Expect(err).ToNot(HaveOccurred())
+
+				if len(pods.Items) > 0 {
+					servicePodIP = pods.Items[0].Status.PodIP
+				}
+
+				return servicePodIP
```
MrFreezeex, line 104: Not sure I am following correctly this test, why are you fetching the pod IP? I am not entirely sure how to test this case shouldn't this get the IP from the ServiceImport somehow?

tpantelis: The service's backend deployment returns the [pod IP](https://github.com/kubernetes-sigs/mcs-api/blob/e2cb6dc0c7538264e2e57cbbe28172ee2dccbb97/conformance/k8s_objects.go#L78) in response to requests. So I check for the pod IP from the deployment on cluster 1 to ensure only this service is accessed.

MrFreezeex: Ahh yes indeed I see it now!

MrFreezeex (approved):
/lgtm

tpantelis:
> /lgtm

I updated the code as I found a simpler way to improve the error reporting using `Eventually` rather than using `AwaitResultOrError`.

Issue #67, opened 2024-09-03
Conformance test: service types


skitt:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md#service-types

Check that:

- [ ] ClusterIP Services produce a ServiceImport of type ClusterSetIP
- [ ] Headless Services produce a ServiceImport of type Headless
- [ ] ExternalName Services produce an InvalidService condition

tpantelis:
@skitt This task can be marked as complete:

- ClusterIP Services produce a ServiceImport of type ClusterSetIP

zhiying-lin:
Looks like Headless Services produce a ServiceImport of type Headless has been addressed in [#91](https://github.com/kubernetes-sigs/mcs-api/pull/91)

skitt:
So it was, thanks! And thanks for taking care of external names.

tpantelis:
`ExternalName Services produce an InvalidService condition` was handled so this PR can be closed.

PR #91, opened 2024-12-11
Add conformance tests for headless ServiceImport

tpantelis:
Also refactored the `ServiceImport` tests a bit. See individual commits for details.

Related to https://github.com/kubernetes-sigs/mcs-api/issues/67

### DIFF
--- a/conformance/service_import.go
+++ b/conformance/service_import.go
@@ -19,7 +19,6 @@ package conformance
 import (
 	"fmt"
 	"net"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -27,7 +26,12 @@ import (
 	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
-var _ = Describe("", func() {
+var (
+	_ = Describe("", testGeneralServiceImport)
+	_ = Describe("", Label(ClusterIPLabel), testClusterIPServiceImport)
+)
+
+func testGeneralServiceImport() {
 	var helloServiceExport *v1alpha1.ServiceExport
 	t := newTestDriver()
 
@@ -49,30 +53,7 @@ var _ = Describe("", func() {
 		}
 	}
 
-	Specify("Exporting a ClusterIP service should create a ServiceImport of type ClusterSetIP in the service's namespace in each cluster. "+
-		"Unexporting should delete the ServiceImport", Label(RequiredLabel, ClusterIPLabel), func() {
-		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#importing-services")
-
-		for i := range clients {
-			serviceImport := t.awaitServiceImport(&clients[i], helloServiceName, nil)
-			Expect(serviceImport).NotTo(BeNil(), reportNonConformant(fmt.Sprintf("ServiceImport was not found on cluster %q",
-				clients[i].name)))
-
-			Expect(serviceImport.Spec.Type).To(Equal(v1alpha1.ClusterSetIP), reportNonConformant(
-				fmt.Sprintf("ServiceImport on cluster %q has type %q", clients[i].name, serviceImport.Spec.Type)))
-		}
-
-		By("Unexporting the service")
-
-		t.deleteServiceExport(&clients[0])
-
-		for i := range clients {
-			t.awaitNoServiceImport(&clients[i], helloServiceName, fmt.Sprintf(
-				"the ServiceImport still exists on cluster %q after unexporting the service", clients[i].name))
-		}
-	})
-
-	// The prior test above also unexports the service and verifies the ServiceImport is deleted, but it doesn't require more than one
+	// Other tests also unexport the service and verifies the ServiceImport is deleted, but it doesn't require more than one
 	// cluster, so it provides basic coverage in a simple single cluster environment. The following test is more comprehensive in that it
 	// exports a service from two clusters and ensures proper behavior when unexported from both.
 	Specify("A ServiceImport should only exist as long as there's at least one exporting cluster", Label(RequiredLabel), func() {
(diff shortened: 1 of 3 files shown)


tpantelis:
/cc @skitt @MrFreezeex 

I didn't add a service port test for headless `ServiceImport` as the spec isn't clear about the expected behavior wrt [computing the union](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port).  It doesn't seem to make sense for an implementation to do so for a headless service. But I don't see this explicitly mentioned in the spec except for an indirect implication in this statement: "_A derived service will be accessible with the **clusterset IP** at the ports dictated by child services_".

conformance/service_import.go
```diff
@@ -214,3 +215,50 @@ func testClusterIPServiceImport() {
...
+	t := newTestDriver()
+
+	BeforeEach(func() {
+		t.helloService.Spec.ClusterIP = corev1.ClusterIPNone
+	})
+
+	JustBeforeEach(func() {
+		t.helloService.Spec.ClusterIP = corev1.ClusterIPNone
+		t.createServiceExport(&clients[0])
+	})
```
MrFreezeex, line 229: Any specific reasons for setting the ClusterIP to None twice?

conformance/service_import.go
```diff
@@ -214,3 +215,50 @@ func testClusterIPServiceImport() {
...
+
+	Specify("No clusterset IP should be allocated for a Headless ServiceImport", Label(RequiredLabel), func() {
+		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#clustersetip")
+
+		serviceImport := t.awaitServiceImport(&clients[0], t.helloService.Name, nil)
+		Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
+
+		Consistently(func() []string {
+			return t.getServiceImport(&clients[0], t.helloService.Name).Spec.IPs
+		}).Within(5*time.Second).ProbeEvery(time.Second).Should(BeEmpty(), reportNonConformant(""))
```
MrFreezeex, line 262: A default service seems to set both `clusterIP` and `clusterIPs` to the `None` string so I think we should do the same in a headless service (also the existing controller does that IIUC.

MrFreezeex:
> I didn't add a service port test for headless `ServiceImport` as the spec isn't clear about the expected behavior wrt [computing the union](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port). It doesn't seem to make sense for an implementation to do so for a headless service. But I don't see this explicitly mentioned in the spec except for an indirect implication in this statement: "_A derived service will be accessible with the **clusterset IP** at the ports dictated by child services_".

There is DNS queries that you could do with port/protocol so this might be influencing that part :man_shrugging:. Although I am not entirely sure we should duplicate a test for that in headless vs clustersetip.

Another good test IMO (not necessarily in this PR and could also do it as you wish) is to try switching from headless to/from clustersetip and check how the implementation would react because it's not an obvious thing to do for implementation with derived service as you can't live update an existing Service ip fields.

conformance/service_import.go
tpantelis: For a `Service`, `clusterIP` is set to `None` by the user to specify that it's headless. I'm not sure why they did it this way rather than a define a service type. However, `ServiceImport` defines the `Headless` type so setting the `ips` field to `[]{"None"}` isn't necessary (and would seem a bit awkward). Also the MCS spec does not say it should be set to `None`. In fact [this section](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-types) in the "_ClusterIP: none (Headless)_" bullet, it says "_Headless services ... do not configure a VIP_" which indicates it should be empty.

> (also the existing controller does that IIUC)

Which controller are you referring to? If it's the reference implementation in this repo, I'm really familiar with the code but it does not explicitly set the `ServiceImport` `ips` field to `None` anywhere. The [ServiceImportReconciler](https://github.com/kubernetes-sigs/mcs-api/blob/8690854b517f61128d8603af11f417692f46930c/pkg/controllers/serviceimport.go#L57) explicitly ignores headless services. OTOH, the [ServiceReconciler](https://github.com/kubernetes-sigs/mcs-api/blob/8690854b517f61128d8603af11f417692f46930c/pkg/controllers/service.go#L71), which sets the `ips` field, does not ignore headless services and blindly copies the `service.Spec.ClusterIP` field which would propagate the `None` value although I think that's an oversight.

tpantelis:
> > I didn't add a service port test for headless `ServiceImport` as the spec isn't clear about the expected behavior wrt [computing the union](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port). It doesn't seem to make sense for an implementation to do so for a headless service. But I don't see this explicitly mentioned in the spec except for an indirect implication in this statement: "_A derived service will be accessible with the **clusterset IP** at the ports dictated by child services_".
> 
> There is DNS queries that you could do with port/protocol so this might be influencing that part 🤷‍♂️. Although I am not entirely sure we should duplicate a test for that in headless vs clustersetip.

I intend to add headless DNS tests later in another PR.

> 
> Another good test IMO (not necessarily in this PR and could also do it as you wish) is to try switching from headless to/from clustersetip and check how the implementation would react because it's not an obvious thing to do for implementation with derived service as you can't live update an existing Service ip fields.

I'm not sure what you mean here. As you said, you can't update a `Service`'s `clusterIP` field after creation (I tried it - it doesn't fail but silently ignores it). So what test would you actually do?

But my original question was should an implementation populate/merge the `ports` field in a `ServiceImport` for a headless service. As I mentioned, it doesn't seem to make sense but I don't see it explicitly excluded in the spec. If it shouldn't then I can add a test to verify the `ports` are empty.

MrFreezeex:
> > Another good test IMO (not necessarily in this PR and could also do it as you wish) is to try switching from headless to/from clustersetip and check how the implementation would react because it's not an obvious thing to do for implementation with derived service as you can't live update an existing Service ip fields.
> 
> I'm not sure what you mean here. As you said, you can't update a `Service`'s `clusterIP` field after creation (I tried it - it doesn't fail but silently ignores it). So what test would you actually do?

For instance, you can have a conflict with a headless/clustersetip and delete the oldest service and it should switch the ServiceImport type.

> But my original question was should an implementation populate/merge the `ports` field in a `ServiceImport` for a headless service. As I mentioned, it doesn't seem to make sense but I don't see it explicitly excluded in the spec. If it shouldn't then I can add a test to verify the `ports` are empty.

That's what I meant with the DNS about protocol and port, afaik coredns will read those and use it to answer those type of queries.

tpantelis:
> > But my original question was should an implementation populate/merge the `ports` field in a `ServiceImport` for a headless service. As I mentioned, it doesn't seem to make sense but I don't see it explicitly excluded in the spec. If it shouldn't then I can add a test to verify the `ports` are empty.
> 
> That's what I meant with the DNS about protocol and port, afaik coredns will read those and use it to answer those type of queries.

By "coredns", I assume you mean https://github.com/coredns/multicluster/? If so, a cursory look at the code, for a headless service, the ports are [extracted from EndpointSlices](https://github.com/coredns/multicluster/blob/3afa45c2996fd02d1d781b74c58da0d0e6855f18/multicluster.go#L351). Submariner does the same thing. The merging semantics of `ServiceImport` `ports` logically only applies to non-headless services and can't be applied to headless services. So there's no point in populating the `ports` for a headless service. the question is...should we enforce this in the conformance tests...

tpantelis:
> For instance, you can have a conflict with a headless/clustersetip and delete the oldest service and it should switch the ServiceImport type.

The spec doesn't say an implementation should switch the service type on the fly in any scenario and I'm not sure it would be valid to do so. The semantics between headless and cluster IP are very different. Imagine clients are using the multi-cluster service's  clusterset VIP and then all of a sudden it switches to headless and the VIP goes away even though the multi-cluster service still exists. Non-MCS services don't allow this but `ServiceImport`s introduce a unique scenario wrt conflict resolutions but I think an MCS implementation should ignore the conflict change and keep the service type as originally determined by conflict resolution, similar to how non-MCS services ignore updates to `clusterIP`.

That said, we can add a test for headless/clusterIP conflict resolution. I can do that in another PR.

MrFreezeex:
> By "coredns", I assume you mean https://github.com/coredns/multicluster/? If so, a cursory look at the code, for a headless service, the ports are [extracted from EndpointSlices](https://github.com/coredns/multicluster/blob/3afa45c2996fd02d1d781b74c58da0d0e6855f18/multicluster.go#L351). Submariner does the same thing. The merging semantics of `ServiceImport` `ports` logically only applies to non-headless services and can't be applied to headless services. So there's no point in populating the `ports` for a headless service. The question is...should we enforce this in the conformance tests? Or should we allow an implementation to use the `ports` field as they see fit for headless services (eg if they don't use `EndpointSlices` for some reason)?

Yes I meant this plugin but I didn't recall that this plugin (or coredns for normal service) did that with EndpointSlice directly, so yes I suppose that there should be nothing that reads those ports when the service is Headless

> > For instance, you can have a conflict with a headless/clustersetip and delete the oldest service and it should switch the ServiceImport type.
> 
> The spec doesn't say an implementation should switch the service type on the fly in any scenario and I'm not sure it would be valid to do so. The semantics between headless and cluster IP are very different. Imagine clients are using the multi-cluster service's clusterset VIP and then all of a sudden it switches to headless and the VIP goes away even though the multi-cluster service still exists. Non-MCS services don't allow this but `ServiceImport`s introduce a unique scenario wrt conflict resolutions but I think an MCS implementation should ignore the conflict change and keep the service type as originally determined by conflict resolution, similar to how non-MCS services ignore updates to `clusterIP`.
> 
> That said, we can add a test for headless/clusterIP conflict resolution. I can do that in another PR. However I'm not sure it would be valid to export a constituent headless service from one cluster together with a clusterIP service from another cluster. Can you just simply treat a constituent headless service as a clusterIP service or vice versa? How should port merging behave? In Submariner, we actually mark the conflicting `ServiceExport` as invalid and don't export it.

Hmm well as the KEP currently doesn't say to exclude endpoints from a service with a different type, maybe it should though :man_shrugging: . And actually it also explicitly state that the type may change dynamically as a result of a conflict in https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#clusterset-service-behavior-expectations.

In cilium we just treat ServiceImport type as it's own property ATM without any effect on other type of conflict (ports, endpoints etc) and the type of the ServiceImport may change if the user change the exported Service type.

tpantelis:
>Hmm well the KEP currently doesn't say to exclude endpoints from a service with a different type, maybe it should though 🤷‍♂️ . And actually it also explicitly state that the type may change dynamically as a result of a conflict in https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#clusterset-service-behavior-expectations.

yeah you mean this (I didn't see this before):

"_Exporting a non-headless service to an otherwise headless service can dynamically change the clusterset service type when an old export is removed, potentially breaking existing consumers. This is likely the result of a deployment error. Conditions and events on the ServiceExport will be used to communicate conflicts to the user._"

I think this implies dynamically changing the service type shouldn't be allowed but the wording is a bit vague. 

> In cilium we just treat ServiceImport type as it's own property ATM without any effect on other properties merging/conflicts (ports, endpoints etc) and the type of the ServiceImport may change if the user change the exported Service type.

The MCS spec doesn't provide clear guidance on how implementations should handle conflicting types other than to apply the general conflict resolution policy, at least currently. But maybe that's intentional to leave it up the implementation. So I think wrt conformance, at this point, we can simply add a test to verify a `Conflict` condition is set. We can expand the test should the spec be updated with further guidance in the future.

tpantelis:
@MrFreezeex It's been a while and I've reviewed the discussions and I don't think there's anything more to do on this PR.

MrFreezeex:
> @MrFreezeex It's been a while and I've reviewed the discussions and I don't think there's anything more to do on this PR.

Ah sorry I thought you were going to fix the reference implementation/ServiceReconciler in https://github.com/kubernetes-sigs/mcs-api/pull/91#discussion_r1882048836 so that it's actually conformant to this new test although by re-reading I thought about an argument for adding None vs no IPs will continue the thread above.

tpantelis:
> Ah sorry I thought you were going to fix the reference implementation/ServiceReconciler in https://github.com/kubernetes-sigs/mcs-api/pull/91#discussion_r1882048836 so that it's actually conformant to this new test although by re-reading I thought about an argument for adding None vs no IPs will continue the thread above.

I didn't intend to change the reference implementation with this PR. in fact I'm not sure I'd even call it a reference implementation b/c it doesn't create ServiceImports which is the basic part of conformance 😄

conformance/service_import.go
MrFreezeex: So by thinking again about this there is actually something else that might be relevant to no IPs vs None. In a derived service style implementation you would do something like this:
- Create the ServiceImport
- Create the derived Service (probably with an owner reference to the ServiceImport, so this order is kinda required)
- Get the Service IP from the derived Service and patch the ServiceImport

Because of this kind of workflow that prevent you to create the ServiceImport with the correct IPs fields from its first creation, an empty IPs fields could mean that the IPs were not initialized yet. But if we don't propagate None this could be because the ServiceImport is headless too.

That being said, I am not strongly against not propagating None/having an empty list if the service is headless as there is not a real ambiguity, it evens means less operation on the apiserver as Headless would not require to patch back the ServiceImport.

MrFreezeex:
> > Ah sorry I thought you were going to fix the reference implementation/ServiceReconciler in [#91 (comment)](https://github.com/kubernetes-sigs/mcs-api/pull/91#discussion_r1882048836) so that it's actually conformant to this new test although by re-reading I thought about an argument for adding None vs no IPs will continue the thread above.
> 
> I didn't intend to change the reference implementation with this PR. in fact I'm not sure I'd even call it a reference implementation b/c it doesn't create ServiceImports which is the basic part of conformance 😄

Well we kinda use it actually: https://github.com/cilium/cilium/blob/main/pkg/clustermesh/mcsapi/cell.go#L109 :sweat_smile:

conformance/service_import.go
tpantelis: As I mentioned above setting to  'None` to indicate headless isn't necessary for `ServiceImport` b/c there's a type field that explicit tells whether it's headless or not. That's the difference with service. Plus setting to `None` is not described in the `ServiceImport` docs nor the KEP.

tpantelis:
> Well we kinda use it actually: https://github.com/cilium/cilium/blob/main/pkg/clustermesh/mcsapi/cell.go#L109 😅

Ahh I didn't know that. I take that back then although it's used as part of a real implementation and is not a complete implementation on its own 😄

conformance/service_import.go
MrFreezeex: My main issue is that the KEP is at best ambiguous on this topic and we don't know if the `ServiceReconciler` propagating `None` is an oversight or intended, from what I can find by reading old PR comments it seems that you are right and an empty list was the reason for the same reasons as you pointed, anyway as I was saying I am not strongly against an empty list. It would be awesome if the `ServiceReconciler` could do the same thing as the conformance test though, happy to do a separate PR if you don't want to do this here though!

tpantelis: I think that should be a separate PR.

MrFreezeex (approved):
/lgtm

tpantelis:
@skitt

MrFreezeex (approved):
/lgtm

skitt (approved):

PR #93, opened 2024-12-19
apis/v1alpha1: add target port to ServiceImport ports

MrFreezeex:
This is adding target port to v1alpha1 API that was missing in the ports definitions. In implementations that relies on derived services this would allow to correctly propagate that information to its derived
Service.

This is also in sync with the current planned change to v1alpha2 which uses directly the core Service ports in its struct definition which already includes target ports.

Kudos to @giorio94 who spotted this!

### DIFF
--- a/pkg/apis/v1alpha1/serviceimport.go
+++ b/pkg/apis/v1alpha1/serviceimport.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // +genclient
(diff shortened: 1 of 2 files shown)


pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -107,6 +108,17 @@ type ServicePort struct {
...
+	// Number or name of the port to access on the pods targeted by the service.
+	// Number must be in the range 1 to 65535. Name must be an IANA_SVC_NAME.
+	// If this is a string, it will be looked up as a named port in the
+	// target Pod's container ports. If this is not specified, the value
+	// of the 'port' field is used (an identity map).
+	// This field is ignored for services with clusterIP=None, and should be
+	// omitted or set equal to the 'port' field.
+	// More info: https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service
+	// +optional
+	TargetPort intstr.IntOrString `json:"targetPort,omitempty"`
```
zhiying-lin, line 121: i asked the same question before in the upstream, [slack discussion](https://kubernetes.slack.com/archives/C09R1PJR3/p1655184999186449)

is it invalid any more?

MrFreezeex: Hmm I didn't see this this thread although, our use case is quite simple: we have derived service in Cilium implementation. If we want the traffic to be routed correctly we need the targetPort property on the derived Service object. Currently targetPort would be the only property that would need us to not reconcile derived services only based of the ServiceImport and this feels weird.

Also ServiceImport was supposed to be able to be implemented (or at least implementable) by kube-proxy at the beginning of the project IIUC or some replacement of kube-proxy. So anything that would attempt to natively support ServiceImport traffic load balancing like a kube-proxy would do on regular services can't do this only via the info available on the API servers currently...

zhiying-lin: It basically depends on the implementation. If it replies on the derived service, the targetPort is needed. Improve the comment to clarify that?

MrFreezeex: I changed a bit the comment above this field to clarify that this field is not only optional for the user but also for the implementation :+1:

MrFreezeex:
After some discussion at the sig meeting, this field wasn't added to keep the API minimalist and while trying to implement it in Cilium it seems to be feasible to pass it down to the derived service without having this on the ServiceImport!

PR #94, opened 2024-12-27
 Update Demo Scripts (make demo) for Apple Silicon (M4) Compatibility and Python 3

kahirokunn:
This pull request updates the “make demo” workflow to ensure that it runs properly on Apple Silicon (M4) machines.

• Switched from Python to Python 3 in the demo script (edit-meta).  
• Used absolute paths to address environment-related issues in demo.sh.  
• Added architecture detection in scripts/up.sh for a multi-arch CoreDNS build.  
• Replaced the tar-based CoreDNS download with an equivalent git clone step to support “make” on M4.

All of these modifications aim to maintain backward compatibility while extending usability to Apple Silicon (M4) hardware. Testing locally shows that “make demo” now works as expected on M4 MacBooks. Please review and let me know if further changes are required. Thank you!

### DIFF
--- a/demo/demo.sh
+++ b/demo/demo.sh
@@ -14,14 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cd $(dirname ${BASH_SOURCE})
-. ./udemo.sh
-. ../scripts/util.sh
+# Resolve the absolute path to this script directory.
+cd $(dirname "${BASH_SOURCE[0]}")
+demo_dir=$(realpath "$(pwd)")
+scripts_dir=$(realpath "$(pwd)/../scripts")
+
+. "${demo_dir}/udemo.sh"
+. "${scripts_dir}/util.sh"
 
 DEMO_AUTO_RUN=true
 
-kubeconfig1=${KUBECONFIG1:-../scripts/c1.kubeconfig}
-kubeconfig2=${KUBECONFIG2:-../scripts/c2.kubeconfig}
+kubeconfig1="${KUBECONFIG1:-${scripts_dir}/c1.kubeconfig}"
+kubeconfig2="${KUBECONFIG2:-${scripts_dir}/c2.kubeconfig}"
 
 k1="kubectl --kubeconfig ${kubeconfig1}"
 k2="kubectl --kubeconfig ${kubeconfig2}"
@@ -30,20 +34,19 @@ desc "Setup our demo namespace"
 run "${k1} create ns demo"
 run "${k2} create ns demo"
 
-c1_pane=`tmux split-window -h -d -P`
+c1_pane=$(tmux split-window -h -d -P)
 
 function cleanup() {
-    tmux kill-pane -t $c1_pane
+    tmux kill-pane -t "$c1_pane"
 }
 trap cleanup EXIT
 
-tmux send -t $c1_pane "${k1} logs -f mcs-api-controller" Enter
+tmux send -t "$c1_pane" "${k1} logs -f mcs-api-controller" Enter
 
 desc "Create our service in each cluster"
-run "${k1} apply -f yaml/dep1.yaml -f yaml/svc.yaml"
-run "${k2} apply -f yaml/dep2.yaml -f yaml/svc.yaml"
+run "${k1} apply -f ${script_dir}/yaml/dep1.yaml -f ${script_dir}/yaml/svc.yaml"
+run "${k2} apply -f ${script_dir}/yaml/dep2.yaml -f ${script_dir}/yaml/svc.yaml"
 run "${k1} get endpointslice -n demo"
-
 
 desc "Lets look at some requests to the service in cluster 1"
 run "${k1} -n demo run -i --rm --restart=Never --image=jeremyot/request:0a40de8 request -- --duration=5s --address=serve.demo.svc.cluster.local"
@@ -51,12 +54,13 @@ run "${k1} -n demo run -i --rm --restart=Never --image=jeremyot/request:0a40de8 
 desc "Ok, looks normal. Let's import the service from our other cluster"
 ep_1=$(${k1} get endpointslice -n demo -l 'kubernetes.io/service-name=serve' --template="{{(index .items 0).metadata.name}}")
 ep_2=$(${k2} get endpointslice -n demo -l 'kubernetes.io/service-name=serve' --template="{{(index .items 0).metadata.name}}")
-run "${k1} get endpointslice -n demo ${ep_1} -o yaml | ./edit-meta --metadata '{name: imported-${ep_1}, namespace: demo, labels: {multicluster.kubernetes.io/service-name: serve}}' > yaml/slice-1.tmp"
-run "${k2} get endpointslice -n demo ${ep_2} -o yaml | ./edit-meta --metadata '{name: imported-${ep_2}, namespace: demo, labels: {multicluster.kubernetes.io/service-name: serve}}' > yaml/slice-2.tmp"
-run "${k1} apply -f yaml/serviceimport.yaml -f yaml/slice-1.tmp -f yaml/slice-2.tmp"
-run "${k2} apply -f yaml/serviceimport.yaml -f yaml/slice-1.tmp -f yaml/slice-2.tmp"
-run "${k1} apply -f yaml/serviceimport-with-vip.yaml -f yaml/slice-1.tmp -f yaml/slice-2.tmp"
-run "${k2} apply -f yaml/serviceimport-with-vip.yaml -f yaml/slice-1.tmp -f yaml/slice-2.tmp"
+
+run "${k1} get endpointslice -n demo ${ep_1} -o yaml | ${script_dir}/edit-meta --metadata '{name: imported-${ep_1}, namespace: demo, labels: {multicluster.kubernetes.io/service-name: serve}}' > ${script_dir}/yaml/slice-1.tmp"
+run "${k2} get endpointslice -n demo ${ep_2} -o yaml | ${script_dir}/edit-meta --metadata '{name: imported-${ep_2}, namespace: demo, labels: {multicluster.kubernetes.io/service-name: serve}}' > ${script_dir}/yaml/slice-2.tmp"
+run "${k1} apply -f ${script_dir}/yaml/serviceimport.yaml -f ${script_dir}/yaml/slice-1.tmp -f ${script_dir}/yaml/slice-2.tmp"
+run "${k2} apply -f ${script_dir}/yaml/serviceimport.yaml -f ${script_dir}/yaml/slice-1.tmp -f ${script_dir}/yaml/slice-2.tmp"
+run "${k1} apply -f ${script_dir}/yaml/serviceimport-with-vip.yaml -f ${script_dir}/yaml/slice-1.tmp -f ${script_dir}/yaml/slice-2.tmp"
+run "${k2} apply -f ${script_dir}/yaml/serviceimport-with-vip.yaml -f ${script_dir}/yaml/slice-1.tmp -f ${script_dir}/yaml/slice-2.tmp"
 
 desc "See what we've created..."
 run "${k1} get -n demo serviceimports"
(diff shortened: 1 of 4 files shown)


demo/demo.sh
```diff
@@ -14,14 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cd $(dirname ${BASH_SOURCE})
-. ./udemo.sh
-. ../scripts/util.sh
+# Resolve the absolute path to this script directory.
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
MrFreezeex, line 18:
```suggestion
cd $(dirname "${BASH_SOURCE[0]}")
script_dir="$(pwd)"
```

demo/demo.sh
```diff
@@ -14,14 +14,18 @@
...
+. "${script_dir}/udemo.sh"
+. "${script_dir}/../scripts/util.sh"
 
 DEMO_AUTO_RUN=true
 
-kubeconfig1=${KUBECONFIG1:-../scripts/c1.kubeconfig}
-kubeconfig2=${KUBECONFIG2:-../scripts/c2.kubeconfig}
+# Use absolute paths for kubeconfig defaults.
+kubeconfig1="${KUBECONFIG1:-$(realpath "${script_dir}/../scripts/c1.kubeconfig")}"
+kubeconfig2="${KUBECONFIG2:-$(realpath "${script_dir}/../scripts/c2.kubeconfig")}"
```
MrFreezeex, line 28: This is already an absolute path, realtpath looks redundant here or am I missing something?

demo/demo.sh
```diff
@@ -14,14 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cd $(dirname ${BASH_SOURCE})
-. ./udemo.sh
-. ../scripts/util.sh
+# Resolve the absolute path to this script directory.
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Source the helper scripts with absolute paths.
```
MrFreezeex, line 20: I don't think we need to re-iterate that we use an absolute path everywhere tbh, if you want you could add that to the first one when script_dir is defined, WDYT?

demo/demo.sh
```diff
@@ -14,14 +14,18 @@
...
+. "${script_dir}/udemo.sh"
+. "${script_dir}/../scripts/util.sh"
 
 DEMO_AUTO_RUN=true
 
-kubeconfig1=${KUBECONFIG1:-../scripts/c1.kubeconfig}
-kubeconfig2=${KUBECONFIG2:-../scripts/c2.kubeconfig}
+# Use absolute paths for kubeconfig defaults.
```
MrFreezeex, line 26: ditto

scripts/up.sh
```diff
@@ -44,15 +65,23 @@ coredns_image="multicluster/coredns:latest"
...
-  tar xf v${coredns_version}.tar.gz
+  # Since the coredns make command uses the git command, we clone it with git
+  git clone --depth 1 https://github.com/coredns/coredns.git --branch v${coredns_version} --single-branch "${coredns_path}"
+
+  # coredns binary is not multi-arch, so we need to download the binary for the current platform
+  platform="$(detect_platform)"
+  coredns_binary_url="https://github.com/coredns/coredns/releases/download/v${coredns_version}/coredns_${coredns_version}_${platform}.tgz"
+  curl -fLO "${coredns_binary_url}"
+  tar xzf "coredns_${coredns_version}_${platform}.tgz"
+  chmod +x coredns
```
MrFreezeex, line 76: The script rebuild the coredns binary because we need to add an out of tree plugin (just a few lines below) so we cannot use the prebuild binary unfortunately

scripts/up.sh
```diff
@@ -44,15 +65,23 @@ coredns_image="multicluster/coredns:latest"
...
   popd
 fi
 pushd "${coredns_path}"
 if ! grep -q -F 'multicluster:github.com/coredns/multicluster' "plugin.cfg"; then
   sed -i -e 's/^kubernetes:kubernetes$/&\nmulticluster:github.com\/coredns\/multicluster/' "plugin.cfg"
 fi
 make
+cp ../coredns ./coredns
```
MrFreezeex, line 84: Ditto, we have to rebuild the coredns binary here

MrFreezeex (changes requested):
Hi! Welcome and thanks for the contribution! I left a couple of comments, most of them are style/nits except the one about downloading the coredns binary instead of building it which I am concerned about but apart from that this looks great thanks again!

/ok-to-test

demo/demo.sh
kahirokunn: Thank you for your feedback. I used `realpath` to prevent paths like `/hoge/fuga/../scripts/c1.kubeconfig` from appearing when using `echo`, aiming for a cleaner output such as `/hoge/scripts/c1.kubeconfig`. However, as you pointed out, it doesn't have any functional impact and is merely a matter of personal preference. I'm not particularly attached to it, so removing `realpath` should be perfectly fine. What are your thoughts on this?

demo/demo.sh
kahirokunn: The proposed commitments have been added. Thank you!

demo/demo.sh
MrFreezeex: Ah yes ok, thanks for the explanation ,no worries then!

scripts/up.sh
kahirokunn: Thank you for your comment.

I reviewed the current `Dockerfile` ([link](https://github.com/coredns/coredns/blob/master/Dockerfile#L14)) and noticed that it primarily utilizes the `docker build` command, which copies the existing CoreDNS binary rather than rebuilding it. This suggests that the binary is not being reconstructed within this specific Docker build process.

Could you please provide more details or point me to the specific script or step where the CoreDNS binary is being rebuilt? Understanding the exact location or method used for the rebuild would greatly help in addressing the integration of the out-of-tree plugin.

I appreciate your guidance and assistance on this matter.

scripts/up.sh
kahirokunn: https://github.com/kubernetes-sigs/mcs-api/pull/94#discussion_r1904575177

🙇

scripts/up.sh
MrFreezeex: Sure that's when `make` is ran right before the `docker build` command: https://github.com/kubernetes-sigs/mcs-api/blob/fede3192824f8c9d44719a467528f9438ae6007b/scripts/up.sh#L55

kahirokunn: I see. Is there a good way to get binaries built that match the container execution environment, rather than binaries that match the host OS environment?
MacOS is hosted on Darwin, but the execution environment is linux.

MrFreezeex: Hmm I see :thinking:, maybe something similar to this then: https://github.com/coredns/coredns?tab=readme-ov-file#compilation-with-docker?

kahirokunn: Thx! 🙏

kahirokunn:
I am currently re-testing the functionality. I will provide an update once it is completed.

kahirokunn:
I've tested this and confirmed it works once this PR is merged.
https://github.com/kubernetes-sigs/mcs-api/pull/97

CleanShot 2025-01-09 at 13 54 50@2x

kahirokunn:
Testing has been completed successfully. 👍

kahirokunn:
@MrFreezeex I think I'm falling down with an error in an unrelated area, what do you think I should do?

scripts/up.sh
```diff
@@ -44,15 +44,16 @@ coredns_image="multicluster/coredns:latest"
...
-  tar xf v${coredns_version}.tar.gz
+  git clone --depth 1 https://github.com/coredns/coredns.git --branch v${coredns_version} --single-branch "${coredns_path}"
   popd
 fi
 pushd "${coredns_path}"
 if ! grep -q -F 'multicluster:github.com/coredns/multicluster' "plugin.cfg"; then
   sed -i -e 's/^kubernetes:kubernetes$/&\nmulticluster:github.com\/coredns\/multicluster/' "plugin.cfg"
 fi
-make
+docker run --rm -i -t \
```
MrFreezeex, line 54: I think your error in the CI is related to the `-i -t` options here

kahirokunn: Thx 🙏

kahirokunn:
It's all been passed!
It's ready to merge. 🙏

MrFreezeex (approved):
Thanks again!
/lgtm

cc @skitt for approval :pray:

demo/demo.sh
```diff
@@ -14,14 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cd $(dirname ${BASH_SOURCE})
-. ./udemo.sh
-. ../scripts/util.sh
+# Resolve the absolute path to this script directory.
+cd $(dirname "${BASH_SOURCE[0]}")
+script_dir="$(pwd)"
```
skitt, line 19: Given that there is a `scripts` directory, but this points to `demo`, it would be less confusing to name this `demo_dir`. It would also be useful to have a pointer to the project root, or the `scripts` directory:

```suggestion
demo_dir=$(realpath "$(pwd)")
scripts_dir=$(realpath "$(pwd)/../scripts")
```

(assignments from command substitutions don’t need to be quoted).

demo/demo.sh
```diff
@@ -14,14 +14,17 @@
...
+
+. "${script_dir}/udemo.sh"
+. "${script_dir}/../scripts/util.sh"
```
skitt, line 22:
```suggestion
. "${demo_dir}/udemo.sh"
. "${scripts_dir}/util.sh"
```

demo/demo.sh
```diff
@@ -14,14 +14,17 @@
...
 
 DEMO_AUTO_RUN=true
 
-kubeconfig1=${KUBECONFIG1:-../scripts/c1.kubeconfig}
-kubeconfig2=${KUBECONFIG2:-../scripts/c2.kubeconfig}
+kubeconfig1="${KUBECONFIG1:-$(realpath "${script_dir}/../scripts/c1.kubeconfig")}"
+kubeconfig2="${KUBECONFIG2:-$(realpath "${script_dir}/../scripts/c2.kubeconfig")}"
```
skitt, line 27:
```suggestion
kubeconfig1="${KUBECONFIG1:-${scripts_dir}/c1.kubeconfig}"
kubeconfig2="${KUBECONFIG2:-${scripts_dir}/c2.kubeconfig}"
```

scripts/up.sh
```diff
@@ -44,15 +44,16 @@ coredns_image="multicluster/coredns:latest"
...
   popd
 fi
 pushd "${coredns_path}"
 if ! grep -q -F 'multicluster:github.com/coredns/multicluster' "plugin.cfg"; then
   sed -i -e 's/^kubernetes:kubernetes$/&\nmulticluster:github.com\/coredns\/multicluster/' "plugin.cfg"
 fi
-make
+docker run --rm \
+    -v $PWD:/go/src/github.com/coredns/coredns -w /go/src/github.com/coredns/coredns \
+        golang:1.23 sh -c 'GOFLAGS="-buildvcs=false" make gen && GOFLAGS="-buildvcs=false" make'
```
skitt, line 56:
```suggestion
        golang:1.23 make gen coredns GOFLAGS=-buildvcs=false
```

skitt (commented):
Thanks for taking care of this! I have a few suggestions for improvements, if you don’t mind.

skitt:
Also, could you rebase your commits to avoid having separate “Apply suggestions from review” commits?

kahirokunn:
@skitt 
Thank you for the review feedback. I have completed rebasing and squashed the separate review suggestion commits. Please review when you have a chance.

kahirokunn:
Bump up 👍

dfarrell07:
> Also, could you rebase your commits to avoid having separate “Apply suggestions from review” commits?

@kahirokunn I suspect Stephen will still want this clean-up of the commit history. Thanks!

kahirokunn:
I see, so I've put everything into one commit.
How about now?

dfarrell07:
> I see, so I've put everything into one commit. How about now?

I suspect that's fine, thanks. The idea was to avoid the "Apply suggestions from code review" commits. I think multiple commits can be great too if the change is more clearly understood/documented that way. Thanks again!

skitt:
Thanks @kahirokunn, as Daniel says my concern was with the “Apply suggestions from review” commits. For future reference it would have been fine to keep different commits for the different changes, as long as the suggestions were merged into the corresponding commit.

/approve

PR #95, opened 2025-01-08
 conformance: fix for dns test with different clusterip on each cluster

MrFreezeex:
In Cilium each ServiceImport can have a different clustersetip on the ServiceImport itself even if the backing pods/endpoints are the same and the service has the same name etc. AFAIK nothing on the spec requires that the IP is the same across all the clusters, this fix the conformance tests to account for those type of implementation.

### DIFF
--- a/conformance/clusterip_service_dns.go
+++ b/conformance/clusterip_service_dns.go
@@ -42,20 +42,20 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, ClusterIPLabel), func() {
 
 		By("Retrieving ServiceImport")
 
-		serviceImport := t.awaitServiceImport(&clients[0], t.helloService.Name, func(serviceImport *v1alpha1.ServiceImport) bool {
-			return len(serviceImport.Spec.IPs) > 0
-		})
-
-		Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
-		Expect(serviceImport.Spec.IPs).ToNot(BeEmpty(), "ServiceImport does not contain an IP")
-
-		clusterSetIP := serviceImport.Spec.IPs[0]
-
-		By(fmt.Sprintf("Found ServiceImport with clusterset IP %q", clusterSetIP))
+		serviceImports := []*v1alpha1.ServiceImport{}
+		for _, client := range clients {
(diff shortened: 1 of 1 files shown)


conformance/clusterip_service_dns.go
```diff
@@ -42,20 +42,20 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, ClusterIPLabel), func() {
...
-		clusterSetIP := serviceImport.Spec.IPs[0]
-
-		By(fmt.Sprintf("Found ServiceImport with clusterset IP %q", clusterSetIP))
+		serviceImports := []*v1alpha1.ServiceImport{}
+		for _, client := range clients {
+			serviceImport := t.awaitServiceImport(&client, t.helloService.Name, func(serviceImport *v1alpha1.ServiceImport) bool {
+				return len(serviceImport.Spec.IPs) > 0
+			})
+			Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
+			Expect(serviceImport.Spec.IPs).ToNot(BeEmpty(), "ServiceImport does not contain an IP")
```
tpantelis, line 51:
```suggestion
			Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found on cluster %q", client.name)
			Expect(serviceImport.Spec.IPs).ToNot(BeEmpty(), "ServiceImport on cluster %q does not contain an IP", client.name)
```

conformance/clusterip_service_dns.go
```diff
@@ -42,20 +42,20 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, ClusterIPLabel), func() {
...
+			serviceImports = append(serviceImports, serviceImport)
+		}
 
 		command := []string{"sh", "-c", fmt.Sprintf("nslookup %s.%s.svc.clusterset.local", t.helloService.Name, t.namespace)}
-
-		for _, client := range clients {
+		for i, client := range clients {
+			clusterSetIP := serviceImports[i].Spec.IPs[0]
+			By(fmt.Sprintf("Found ServiceImport with clusterset IP %q", clusterSetIP))
```
tpantelis, line 58:
```suggestion
			By(fmt.Sprintf("Found ServiceImport on cluster %q with clusterset IP %q", client.name, clusterSetIP))
```

tpantelis (changes requested):

PR #96, opened 2025-01-08
conformance: fix annotations/labels test to not be exact match

MrFreezeex:
This relax the test since the spec about annotations/labels sync does say that labels/annotations should not be taken from the Service exported but it doesn't says that those should be the only annotation/labels ever present on the ServiceImport object. For instance in Cilium we use this annotation: https://github.com/kubernetes-sigs/mcs-api/blob/fede3192824f8c9d44719a467528f9438ae6007b/pkg/controllers/common.go#L35.

### DIFF
--- a/conformance/service_import.go
+++ b/conformance/service_import.go
@@ -38,6 +38,17 @@ var _ = Describe("", func() {
 		t.createServiceExport(&clients[0], helloServiceExport)
 	})
 
+	assertHasKeyValues := func(actual, expected map[string]string) {
+		for k, v := range expected {
+			Expect(actual).To(HaveKeyWithValue(k, v), reportNonConformant(""))
+		}
+	}
+	assertNotHasKeyValues := func(actual, expected map[string]string) {
+		for k, v := range expected {
+			Expect(actual).ToNot(HaveKeyWithValue(k, v), reportNonConformant(""))
+		}
+	}
+
 	Specify("Exporting a ClusterIP service should create a ServiceImport of type ClusterSetIP in the service's namespace in each cluster. "+
 		"Unexporting should delete the ServiceImport",
 		Label(RequiredLabel), func() {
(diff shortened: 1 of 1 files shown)


conformance/service_import.go
```diff
@@ -231,8 +231,11 @@ var _ = Describe("", func() {
 					})
 					Expect(serviceImport).NotTo(BeNil(), "ServiceImport was not found")
 
-					Expect(serviceImport.Annotations).To(Equal(helloServiceExport.Spec.ExportedAnnotations), reportNonConformant(""))
-					Expect(serviceImport.Labels).To(Equal(helloServiceExport.Spec.ExportedLabels), reportNonConformant(""))
+					Expect(serviceImport.Annotations).To(HaveKeyWithValue("dummy-annotation", "true"), reportNonConformant(""))
+					Expect(serviceImport.Annotations).To(Not(ContainElements("dummy-annotation2")), reportNonConformant(""))
+
```
tpantelis, line 236: I think it would be cleaner to generalize these assertions:
                 
```
 assertHasKeyValues := func(actual, expected map[string]string) {
    for k, v := range expected {
        Expect(actual).To(HaveKeyWithValue(k, v), reportNonConformant(""))
    }
}

assertNotHasKeyValues := func(actual, expected map[string]string) {
    for k, v := range expected {
        Expect(actual).ToNot(HaveKeyWithValue(k, v), reportNonConformant(""))
    }
}
...
			
assertHasKeyValues(serviceImport.Annotations, helloServiceExport.Spec.ExportedAnnotations)

assertNotHasKeyValues(serviceImport.Annotations, helloServiceExport2.Spec.ExportedAnnotations)
```
		
Same elsewhere.

tpantelis (changes requested):

conformance/service_import.go
MrFreezeex: Thanks! indeed makes sense, I just did this change let me know if there's anything else!

PR #98, opened 2025-01-10
controllers: fix ServiceReconciler headless service reconciliation

MrFreezeex:
Headless services should have an empty list of IPs instead of a list containing the "None" string.

### DIFF
--- a/pkg/controllers/service.go
+++ b/pkg/controllers/service.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"slices"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


MrFreezeex:
This should be good too, I rebased it after #80 merge. When you have a moment to also review this @skitt @tpantelis that would be awesome :pray: (and thanks for all the other PRs that you already reviewed btw :heart:).

FYI I would like to bump mcs-api in our coming days as a "last minute fix" to make it in time for backporting it to Cilium next stable release  :pray:. There is still like probably a full week for that I don't  need it absolutely now there is still a bit of time for my constraints here!

PR #99, opened 2025-01-13
Conformance: verify ServiceImport doesn't contain Service labels/annotations

tpantelis:
...since labels/annotations should only be taken from the ServiceExport.


MrFreezeex (approved):
Thanks!
/lgtm

PR #100, opened 2025-01-24
Bump to coredns 1.11.4 for the multicluster plugin

skitt:
The multicluster plugin requires 1.11.4. This remains fragile since the container build will always pick up the latest multicluster plugin...


MrFreezeex (approved):
Thanks!
/lgtm

Also fyi I am still trying to have the multicluster plugin merged in-tree: https://github.com/coredns/coredns/pull/7082 which should also help version management wise

PR #101, opened 2025-01-24
Split conformance, controllers and e2e into separate modules

skitt:
This simplifies the main go.mod, reducing the impact on downstreams.


conformance/go.mod
```diff
@@ -0,0 +1,60 @@
...
+	github.com/onsi/ginkgo/v2 v2.17.1
+	github.com/onsi/gomega v1.33.0
+	k8s.io/api v0.30.0
+	k8s.io/apimachinery v0.30.0
+	k8s.io/client-go v0.30.0
+	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
+	sigs.k8s.io/mcs-api v0.1.0
+)
+
+replace sigs.k8s.io/mcs-api => ..
```
tpantelis, line 15: @skitt How will this affect users that import `sigs.k8s.io/mcs-api/conformance` via Go?

skitt: `replace` directives only apply to the main module, so it will be ignored by users importing the `conformance` module. But we will have to be careful when releasing — the `mcs-api` import will need to be updated.

PR #103, opened 2025-01-29
Add conformance tests for headless DNS

tpantelis:
Verify type A and SRV records are returned from DNS.

Verify individual pods are accessible through "_hostname.clusterid.svc.ns.svc.clusterset.local_" for a StatefulSet service, according to this section from the spec:

"_Pods backing a Headless service may be addressed individually using the `<hostname>.<clusterid>.<svc>.<ns>.svc.clusterset.local` format; necessary records will be created based on each ready endpoint's hostname and the multicluster.kubernetes.io/source-cluster label on the EndpointSlice. This allows naming collisions to be avoided for headless services backed by identical StatefulSets deployed in multiple clusters._"

Related to https://github.com/kubernetes-sigs/mcs-api/issues/69

### DIFF
--- a/conformance/endpoint_slice.go
+++ b/conformance/endpoint_slice.go
@@ -17,7 +17,6 @@ limitations under the License.
 package conformance
 
 import (
-	"context"
 	"fmt"
 	"time"
 
@@ -26,7 +25,6 @@ import (
 	discoveryv1 "k8s.io/api/discovery/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 )
 
@@ -47,23 +45,18 @@ var _ = Describe("", Label(OptionalLabel, EndpointSliceLabel), func() {
 		endpointSlices := make([]*discoveryv1.EndpointSlice, len(clients))
 
 		for i, client := range clients {
-			eps := t.awaitMCSEndpointSlice(&client)
-			Expect(eps).ToNot(BeNil(), reportNonConformant(fmt.Sprintf(
+			eps := t.awaitMCSEndpointSlice(&client, discoveryv1.AddressTypeIPv4, nil, reportNonConformant(fmt.Sprintf(
 				"an MCS EndpointSlice was not found on cluster %q. An MCS EndpointSlice is identified by the presence "+
-					"of at least one of the required MCS labels, whose names are prefixed with \"multicluster.kubernetes.io\". "+
+					"of the required MCS labels (%q and %q). "+
 					"If the MCS implementation does not use MCS EndpointSlices, you can specify a Ginkgo label filter using "+
 					"the %q label where appropriate to skip this test.",
-				client.name, EndpointSliceLabel)))
+				client.name, v1alpha1.LabelServiceName, v1alpha1.LabelSourceCluster, EndpointSliceLabel)))
 
 			endpointSlices[i] = eps
 
 			Expect(eps.Labels).To(HaveKeyWithValue(v1alpha1.LabelServiceName, t.helloService.Name),
 				reportNonConformant(fmt.Sprintf("the MCS EndpointSlice %q does not contain the %q label referencing the service name",
 					eps.Name, v1alpha1.LabelServiceName)))
-
-			Expect(eps.Labels).To(HaveKey(v1alpha1.LabelSourceCluster),
-				reportNonConformant(fmt.Sprintf("the MCS EndpointSlice %q does not contain the %q label",
-					eps.Name, v1alpha1.LabelSourceCluster)))
 
 			Expect(eps.Labels).To(HaveKey(discoveryv1.LabelManagedBy),
 				reportNonConformant(fmt.Sprintf("the MCS EndpointSlice %q does not contain the %q label",
@@ -90,7 +83,8 @@ var _ = Describe("", Label(OptionalLabel, EndpointSliceLabel), func() {
 	})
 })
 
-func (t *testDriver) awaitMCSEndpointSlice(c *clusterClients) *discoveryv1.EndpointSlice {
+func (t *testDriver) awaitMCSEndpointSlice(c *clusterClients, addressType discoveryv1.AddressType,
+	verify func(Gomega, *discoveryv1.EndpointSlice), desc ...any) *discoveryv1.EndpointSlice {
 	var endpointSlice *discoveryv1.EndpointSlice
 
 	hasLabel := func(eps *discoveryv1.EndpointSlice, label string) bool {
@@ -98,24 +92,29 @@ func (t *testDriver) awaitMCSEndpointSlice(c *clusterClients) *discoveryv1.Endpo
 		return exists
 	}
 
-	_ = wait.PollUntilContextTimeout(ctx, 100*time.Millisecond,
-		20*time.Second, true, func(ctx context.Context) (bool, error) {
-			defer GinkgoRecover()
+	Eventually(func(g Gomega) {
+		list, err := c.k8s.DiscoveryV1().EndpointSlices(t.namespace).List(ctx, metav1.ListOptions{})
+		g.Expect(err).ToNot(HaveOccurred(), "Error retrieving EndpointSlices")
 
-			list, err := c.k8s.DiscoveryV1().EndpointSlices(t.namespace).List(ctx, metav1.ListOptions{})
-			Expect(err).ToNot(HaveOccurred(), "Error retrieving EndpointSlices")
+		endpointSlice = nil
 
-			for i := range list.Items {
-				eps := &list.Items[i]
+		for i := range list.Items {
+			eps := &list.Items[i]
 
-				if hasLabel(eps, v1alpha1.LabelServiceName) || hasLabel(eps, v1alpha1.LabelSourceCluster) {
-					endpointSlice = eps
-					return true, nil
+			if hasLabel(eps, v1alpha1.LabelServiceName) && hasLabel(eps, v1alpha1.LabelSourceCluster) && eps.AddressType == addressType {
+				endpointSlice = eps
+
+				if verify != nil {
+					verify(g, endpointSlice)
 				}
 			}
+		}
 
-			return false, nil
-		})
+		g.Expect(endpointSlice).ToNot(BeNil(), desc...)
+
+		// The final run succeeded so cancel any prior non-conformance reported.
+		cancelNonConformanceReport()
+	}).Within(20 * time.Second).ProbeEvery(100 * time.Millisecond).Should(Succeed())
 
 	return endpointSlice
 }
--- a/conformance/headless_service_dns.go
+++ b/conformance/headless_service_dns.go
@@ -1,0 +1,264 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conformance
+
+import (
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
+	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
+)
+
+var _ = Describe("", Label(OptionalLabel, DNSLabel, HeadlessLabel), func() {
+	const replicas = 2
+
+	t := newTestDriver()
+
+	BeforeEach(func() {
+		t.helloService.Spec.ClusterIP = corev1.ClusterIPNone
+		t.helloDeployment.Spec.Replicas = ptr.To(int32(replicas))
+	})
+
+	JustBeforeEach(func() {
+		t.createServiceExport(&clients[0], newHelloServiceExport())
+	})
+
+	Specify("A DNS query of the <service>.<ns>.svc.clusterset.local domain for a headless service should return the "+
+		"ready endpoint addresses of all the backing pods", func() {
+		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns")
+
+		command := []string{"sh", "-c", fmt.Sprintf("nslookup %s.%s.svc.clusterset.local", t.helloService.Name, t.namespace)}
+
+		endpoints := t.awaitK8sEndpoints(&clients[0], discovery.AddressTypeIPv4)
+
+		var addresses []string
+		for _, ep := range endpoints {
+			addresses = append(addresses, ep.address)
+		}
+
+		for _, client := range clients {
+			By(fmt.Sprintf("Executing command %q on cluster %q", strings.Join(command, " "), client.name))
+
+			t.awaitCmdOutputMatches(&client, command, HaveAddresses(addresses), 1, reportNonConformant(""))
+		}
+	})
+
+	Context("", func() {
+		BeforeEach(func() {
+			t.helloDeployment = nil
+		})
+
+		JustBeforeEach(func() {
+			_, err := clients[0].k8s.AppsV1().StatefulSets(t.namespace).Create(ctx, newStatefulSet(replicas), metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Specify("A DNS query of the <hostname>.<clusterid>.<service>.<ns>.svc.clusterset.local domain for a headless StatefulSet "+
+			"service should return the requested pod's endpoint address", Label(EndpointSliceLabel), func() {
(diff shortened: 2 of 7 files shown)


conformance/headless_service_dns.go
```diff
@@ -0,0 +1,216 @@
...
+					continue
+				}
+
+				epi := endpointInfo{}
+
+				switch {
+				case ptr.Deref(ep.Hostname, "") != "":
+					epi.hostName = *ep.Hostname
+				case ep.TargetRef != nil:
+					epi.hostName = ep.TargetRef.Name
```
MrFreezeex, line 153: Ah I wasn't aware that DNS was using target ref as a fallback so I am absolutely not doing that in Cilium right now :/. Kudos to you then, I have something to fix in Cilium!

Which also makes me think, I guess we should fail "early" on empty hostname perhaps?

MrFreezeex: And even check  that the hostname between clusters match across clusters, although this is maybe more something for a test in `endpoint_slice.go` / possibly not in this PR even

tpantelis: I'm not sure I follow what you mean. TBH, I grabbed this code snippet from Submariner E2E code and I didn't write it 😄 Perhaps we could get the pod IPs and host names from the Pod resources themselves instead of the EndpointSlices.

tpantelis: > Ah I wasn't aware that DNS was using target ref as a fallback

Looking at an EndpointSlice for a headless service I created:

```
- Addresses:  10.244.0.8
    Conditions:
      Ready:    true
    Hostname:   <unset>
    TargetRef:  Pod/nginx-demo-569f89c9f4-q8ddw
    NodeName:   cluster1
```

Only the `TargetRef` is set and references the pod name. I'm not sure when `Hostname` would be set.

MrFreezeex: Hostname is essentially set on the governing services of a statefulset.

Actually reading this I thought that CoreDNS (with the normal kubernetes plugin) handled `TargetRef` as a hostname fallback but it actually does that only when the option `endpoint_pod_names` is enabled (disabled by default): https://github.com/coredns/coredns/blob/d8ecde1080e7cbbeb98257ba4e03a271f16b4cd9/plugin/kubernetes/kubernetes.go#L331

So this might not be right but I interpreted in the mcs KEP "hostname" as only the actualy hostname field not also the pod name :sweat_smile:. We actually have gibberish name in `TargetRef` when syncing across clusters (`fake-pod-10.xx.xx.xx` IIRC). I am not sure what was intended though as as stated in the previous paragraph CoreDNS does not do that by default for regular services :thinking:

MrFreezeex: Well it depends, if the KEP when talking about `<hostname>` means the field `Hostname` or the backing pod name this would be the correct check to do I think.

Else if `<hostname>` is actually supposed ot be only the `Hostname`  field afaik to do this test we would essentially have to create a Statefulset with setting a governing service to the headless service (setting `spec.serviceName` essentially) or crafting the EndpointSlice manually and setting the `Hostname` field when doing that I guess.

tpantelis: Looking at the spec again:

"_Pods backing a Headless service may be addressed individually using the <hostname>.<clusterid>.<svc>.<ns>.svc.clusterset.local format; necessary records will be created based on each ready endpoint's hostname and the multicluster.kubernetes.io/source-cluster label on the EndpointSlice. This allows naming collisions to be avoided for headless services backed by identical StatefulSets deployed in multiple clusters._"

Now I see what you mean. So if I understand this correctly, the  `<hostname>.<clusterid>.<svc>...` format is intended for a headless service backed by a StatefulSet and the host name and  cluster ID is intended to be obtained from the MCS EndpointSlice.

tpantelis: I created a `StatefulSet` and service and the resulting `EndpointSlice` was:

```
- Addresses:  10.244.0.117
    Conditions:
      Ready:    true
    Hostname:   web-1
    TargetRef:  Pod/web-1
    NodeName:   cluster1
```

So `Hostname` is set in this case.

tpantelis: Designing the test in this manner makes more sense and alleviates the issue with determining the cluster ID that I  mentioned as a caveat in PR desc. I should've read the spec in more detail before doing this test 😄

MrFreezeex: No worries!

And actually I don't think you can really take the `NodeName` for determining the cluster id, to me you would essentially have the same issue except if you start reading the EndpointSlice label synced but I am not sure if it actually makes sense to do that.

tpantelis: > I don't think you can really take the NodeName for determining the cluster id

No I was using the value extracted from the kube context which isn't reliable.

I would have to get the cluster ID from the MCS EndpointSlice label as the spec says or have it specified as a CLI arg. I think the former is reasonable - the test would rely on the optional MCS EndpointSlice but that's OK - we can skip it if there's no EndpointSlice.

MrFreezeex:
Hi! First of all thanks a lot again for these new conformances tests, this helped me spotted a bug with the new version of the multicluster coredns plugin!

I have checked against Cilium and I think one of us got something wrong still with the hostname as I got these errors running those new tests:
```
• [FAILED] [28.081 seconds]
 [It] A DNS query of the <hostname>.<clusterid>.<service>.<ns>.svc.clusterset.local domain for a headless service should return the requested pod's endpoint address [Optional, DNS, Headless]
/home/aouthenin/mcs-api/conformance/headless_service_dns.go:70

  Timeline >>
  spec-ref - /home/aouthenin/mcs-api/conformance/headless_service_dns.go:72 @ 02/14/25 23:54:11.381
    https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns
  STEP: Retrieving endpoint addresses for the service on cluster "kind-clustermesh1" @ 02/14/25 23:54:11.381
  STEP: Found endpoints [address:"10.1.0.184", hostName:"hello-77d4cc6984-r4p7d" address:"10.1.1.57", hostName:"hello-77d4cc6984-52tnr"] @ 02/14/25 23:54:11.383
  STEP: Executing command "sh -c nslookup hello-77d4cc6984-r4p7d.kind-clustermesh1.hello.mcs-conformance-675109094.svc.clusterset.local" on cluster "kind-clustermesh1" @ 02/14/25 23:54:11.383
  non-conformant - /home/aouthenin/mcs-api/conformance/report.go:204 @ 02/14/25 23:54:31.383
  [FAILED] in [It] - /home/aouthenin/mcs-api/conformance/conformance_suite.go:313 @ 02/14/25 23:54:31.383
  << Timeline

  [FAILED] Timed out after 20.000s.

  The function passed to Eventually failed at /home/aouthenin/mcs-api/conformance/conformance_suite.go:312 with:
  Command output
  Expected
      <string>:
  to have addresses
      <[]string | len:1, cap:1>: ["10.1.0.184"]
  In [It] at: /home/aouthenin/mcs-api/conformance/conformance_suite.go:313 @ 02/14/25 23:54:31.383
------------------------------
• [FAILED] [8.110 seconds]
 [It] A DNS SRV query of the <service>.<ns>.svc.clusterset.local domain for a headless service should return valid SRV records [Optional, DNS, Headless]
/home/aouthenin/mcs-api/conformance/headless_service_dns.go:87

  Timeline >>
  spec-ref - /home/aouthenin/mcs-api/conformance/headless_service_dns.go:89 @ 02/14/25 23:54:39.451
    https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns
  STEP: Retrieving endpoint addresses for the service on cluster "kind-clustermesh1" @ 02/14/25 23:54:39.451
  STEP: Found endpoints [address:"10.1.0.167", hostName:"hello-77d4cc6984-qlpcj" address:"10.1.1.168", hostName:"hello-77d4cc6984-gs9mr"] @ 02/14/25 23:54:39.453
  STEP: Executing command "sh -c nslookup -type=SRV hello.mcs-conformance-747615969.svc.clusterset.local" on cluster "kind-clustermesh1" @ 02/14/25 23:54:39.453
  non-conformant - /home/aouthenin/mcs-api/conformance/report.go:204 @ 02/14/25 23:54:39.494
    SRV record for endpoint host name "hello-77d4cc6984-qlpcj" not received. Actual records received: [port:42, domainName:"10-1-1-168.clustermesh1.hello.mcs-conformance-747615969.svc.clusterset.local" port:42, domainName:"10-1-0-167.clustermesh1.hello.mcs-conformance-747615969.svc.clusterset.local"]
  [FAILED] in [It] - /home/aouthenin/mcs-api/conformance/headless_service_dns.go:106 @ 02/14/25 23:54:39.494
  << Timeline

  [FAILED] SRV record for endpoint host name "hello-77d4cc6984-qlpcj" not received. Actual records received: [port:42, domainName:"10-1-1-168.clustermesh1.hello.mcs-conformance-747615969.svc.clusterset.local" port:42, domainName:"10-1-0-167.clustermesh1.hello.mcs-conformance-747615969.svc.clusterset.local"]
  Expected
      <int>: -1
  to be >=
      <int>: 0
  In [It] at: /home/aouthenin/mcs-api/conformance/headless_service_dns.go:106 @ 02/14/25 23:54:39.494
------------------------------
```


As a small example with one service exported by two clusters each having two pods I got the following output on a SRV request (+ "A" query to one of the entry):
```
tmp-shell:~# dig SRV rebel-base-headless-mcsapi.default.svc.clusterset.local

; <<>> DiG 9.18.25 <<>> SRV rebel-base-headless-mcsapi.default.svc.clusterset.local
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 62604
;; flags: qr aa rd; QUERY: 1, ANSWER: 6, AUTHORITY: 0, ADDITIONAL: 7
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 00fa4a13ba7eda8e (echoed)
;; QUESTION SECTION:
;rebel-base-headless-mcsapi.default.svc.clusterset.local. IN SRV

;; ANSWER SECTION:
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 fd00-10-1-1--a084.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 fd00-10-1--5c3c.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 10-1-0-73.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 10-2-1-246.clustermesh2.rebel-base-headless-mcsapi.default.svc.clusterset.local.
rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN SRV 0 16 80 10-2-0-83.clustermesh2.rebel-base-headless-mcsapi.default.svc.clusterset.local.

;; ADDITIONAL SECTION:
10-2-1-246.clustermesh2.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN A 10.2.1.246
fd00-10-1--5c3c.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN AAAA fd00:10:1::5c3c
10-2-0-83.clustermesh2.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN A 10.2.0.83
10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN A 10.1.1.220
10-1-0-73.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN A 10.1.0.73
fd00-10-1-1--a084.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN AAAA fd00:10:1:1::a084

;; Query time: 0 msec
;; SERVER: 172.20.1.10#53(172.20.1.10) (UDP)
;; WHEN: Fri Feb 14 16:49:19 UTC 2025
;; MSG SIZE  rcvd: 820

tmp-shell:~# dig A 10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.

; <<>> DiG 9.18.25 <<>> A 10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local.
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 12934
;; flags: qr aa rd; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: fb02ebac44301388 (echoed)
;; QUESTION SECTION:
;10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. IN A

;; ANSWER SECTION:
10-1-1-220.clustermesh1.rebel-base-headless-mcsapi.default.svc.clusterset.local. 5 IN A 10.1.1.220

;; Query time: 0 msec
;; SERVER: 172.20.1.10#53(172.20.1.10) (UDP)
;; WHEN: Fri Feb 14 16:49:30 UTC 2025
;; MSG SIZE  rcvd: 215
```

So right now I don't have any hostname and it seems that coredns is essentially getting the ip address as hostname, also this seems consistent with similar queries to the `cluster.local` domain.

tpantelis:
< So right now I don't have any hostname and it seems that coredns is essentially getting the ip address as hostname, also this seems consistent with similar queries to the cluster.local domain.

You are correct - both the [K8s CoreDNS plugin](https://github.com/coredns/coredns/blob/669ff527bf5275b34d4a3626642e9bddf26ad1ae/plugin/kubernetes/kubernetes.go#L327) and the [multi-cluster plugin](https://github.com/coredns/multicluster/blob/49f47d950355f793d656aec8a6d198daf1d888b1/multicluster.go#L401) fallback to the dashed form of the endpoint IP address if the host name field isn't set. The Submariner plugin falls back to the `TargetRef` name, if set, hence why I checked for it in the tests. It was added several years ago by another dev for stateful sets in case `Hostname` wasn't set, tho not sure if that case is valid now. Either way this behavior is inconsistent with K8s and I've created an issue in Submariner to address it.

skitt:
Does this still need updating?

tpantelis:
> Does this still need updating?

Yes - I'll do it when I get some cycles to focus on it.

tpantelis:
@skitt @MrFreezeex OK - I made the change as outlined in https://github.com/kubernetes-sigs/mcs-api/pull/103#issuecomment-2677153864.

MrFreezeex (approved):
I unfortunately cannot fully test it because I discovered a bug in Cilium preventing us to sync hostname in EndpointSlice for MCS-API but the code and the test behavior lgtm!

tpantelis:
> I unfortunately cannot fully test it because I discovered a bug in Cilium preventing us to sync hostname in EndpointSlice for MCS-API but the code and the test behavior lgtm!

I tested it with Submariner and it works fine. One test fails but it's expected b/c Submariner returns the `TargetRef` name (I haven't addressed that issue yet in Submariner).

conformance/endpoint_slice.go
```diff
@@ -90,32 +83,37 @@ var _ = Describe("", Label(OptionalLabel, EndpointSliceLabel), func() {
...
-				eps := &list.Items[i]
+			if hasLabel(eps, v1alpha1.LabelServiceName) && hasLabel(eps, v1alpha1.LabelSourceCluster) {
+				endpointSlice = eps
 
-				if hasLabel(eps, v1alpha1.LabelServiceName) || hasLabel(eps, v1alpha1.LabelSourceCluster) {
-					endpointSlice = eps
-					return true, nil
+				if verify != nil {
+					verify(g, endpointSlice)
 				}
```
MrFreezeex, line 108: Argh I spent a little while fixing Cilium and I noticed a pretty niche and annoying situation that can break this test. In regular kubernetes Services if you have pods with only ipv4 addresses and the service is dual stack the EndpointSlice controller will create an empty EndpointSlice for ipv6 and an EndpointSlice for ipv4 that contains something.

Now applied to MCS-API this essentially means that if cluster1 is ipv4 only and cluster2 is dual stack you could have an empty EndpointSlice for ipv6 in cluster2 too and that's what Cilium does essentially.

I am not entirely sure what would be the best course of action here :/, maybe ignore empty EndpointSlice? WDYT?

tpantelis: So I assume you're seeing test failures with Cilium. If so which one(s)? Is there an issue with code changes in this PR or in general? If the latter then I propose we merge this PR and address your issue in another.

So it sounds like Cilium creates an empty ipv6 `EndpointSlice` for an ipv4-only service imported from another cluster. If so, then it seems we should ignore any `EndpointSlice` that doesn't match the service's IP family.

MrFreezeex: > So I assume you're seeing test failures with Cilium. If so which one(s)? Is there an issue with code changes in this PR or in general? If the latter then I propose we merge this PR and address your issue in another.

Only in this new statefulset test as the other EndpointSlice related test where `awaitMCSEndpointSlice` is called I didn't properly integrated that into Cilium yet (as we do not create an EndpointSlice for a non headless service by default). The new way this function is written means that when this dualstack/non dualstack situation happens it fails every-time with the way ` awaitMCSEndpointSlice` as it now does the check for every EndpointSlice with matching labels vs before only the first one (this is probably not a bad thing though).

I don't mind that we fix this here or in another PR though.

> So it sounds like Cilium creates an empty ipv6 EndpointSlice for an ipv4-only service imported from another cluster. If so, then it seems we should ignore any EndpointSlice that doesn't match the service's IP family.

Hmm I guess that would be possible to it means that in those tests that looks at the exported Service ip family on clusters that have pods (typically first cluster here) and depending on this ignore the EndpointSlice not matching the ip family on the rest of the clusters/"consumers only" cluster. Is that what you mean?

tpantelis: Yes. The test creates the exported service so is in control of the IP family. So if a test creates and exports an IPv4 service then only look for `EndpointSlice`(s) whose `AddressType` is "IPv4" and, similarly for IPv6 when tests for that are added at some point.

MrFreezeex: Thanks a lot, I can confirm we are passing the test now :pray:

PR #105, opened 2025-05-21
apis: add short names to crds

MrFreezeex:
Add svcexport/svcimport short name that might feel more natural to type as the short name of service is svc. Also preserve the existing default short name "svcex" and "svcim".


MrFreezeex:
I need to check because without updating the crd update command it doesn't actually generate the new manifest :/

MrFreezeex:
I made it so that we don't have to nuke/change the CRD generation process :D!

Those field were actually controlled by crd-base mainly but IMO it's still a good thing to add those as kubebuilder tag so that if someday we remove the crd-base manifests we would not loose these short names

Issue #92, opened 2024-12-12
Conformance test: verify resolution of conflicting service types


tpantelis:
Re: https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#headlessness

Add a test that first exports a ClusterIP service from cluster1 then exports a headless service from cluster2 and `ServiceExport` `Conflict` condition is set and the resulting `ServiceImport` type is `ClusterSetIP`.

PR #111, opened 2025-06-23
Add conformance test to verify resolution of conflicting service types

tpantelis:
Also refactored/improved `awaitServiceImport` in a separate commit.

Fixes https://github.com/kubernetes-sigs/mcs-api/issues/92

### DIFF
--- a/conformance/service_import.go
+++ b/conformance/service_import.go
@@ -47,14 +47,15 @@ func testGeneralServiceImport() {
 		t.createServiceExport(&clients[0], helloServiceExport)
 	})
 
-	assertHasKeyValues := func(actual, expected map[string]string) {
+	assertHasKeyValues := func(g Gomega, actual, expected map[string]string) {
 		for k, v := range expected {
-			Expect(actual).To(HaveKeyWithValue(k, v), reportNonConformant(""))
+			g.Expect(actual).To(HaveKeyWithValue(k, v), reportNonConformant(""))
 		}
 	}
-	assertNotHasKeyValues := func(actual, expected map[string]string) {
+
+	assertNotHasKeyValues := func(g Gomega, actual, expected map[string]string) {
 		for k, v := range expected {
-			Expect(actual).ToNot(HaveKeyWithValue(k, v), reportNonConformant(""))
+			g.Expect(actual).ToNot(HaveKeyWithValue(k, v), reportNonConformant(""))
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


conformance/service_import.go
```diff
@@ -326,3 +327,29 @@ func testExternalNameService() {
...
+
+	JustBeforeEach(func() {
+		t.createServiceExport(&clients[0], newHelloServiceExport())
+	})
+
+	Specify("A service exported on two clusters with conflicting headlessness should apply the conflict resolution policy and "+
+		"report a Conflict condition on the ServiceExport", Label(RequiredLabel), func() {
+		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#headlessness")
+
+		t.awaitServiceExportCondition(&clients[1], v1alpha1.ServiceExportConflict, metav1.ConditionTrue)
```
MrFreezeex, line 346: BTW shouldn't this be done on both clusters?

MrFreezeex: The KEP says this for reference:
```
Conflict resolution policy: If any properties have conflicting values that can not simply be merged, a ServiceExportConflict condition will be set on all ServiceExports for the conflicted service with a description of the conflict.
```

MrFreezeex: (and it's also what we are doing elsewhere it seems)

tpantelis: I know but that's problematic unless every cluster has access to the `ServiceExports` on every other cluster or there's a central controller that has access to all. Submariner has neither. This is an assumption that the KEP makes that it really shouldn't that's been discussed in the past. Plus I don't think it really needs to be on every `ServiceExport` in this case. The important thing is that it's on the `ServiceExport` in the cluster that is actually in conflict.

MrFreezeex: Ah yes ok I see, well to me the conformance test should have consistent tests so just for this matter I think it should have the test for both cluster because I don't think the other conflict tests are any different.

That being said I would be supportive of loosing the tone on the KEP to accommodate your need (for instance saying that implementation are required to have a conflict condition on the "loosing export" but that it's recommended to have it on every ServiceExport if feasible) and to change all test doing similar check once merged.

tpantelis: The only way I can see for Submariner or a similar implementation to accomplish setting a Conflict condition on every exporting cluster is if the contributing information from each cluster is stored in the `ServiceImport` status, including the `ServiceImport` spec fields and `ServiceExport` creation date. Right now only the cluster name is stored. This way, each cluster could independently evaluate for conflicts and report any conflicts on their local `ServiceExport`.

zhiying-lin: i thought we decided to relax the statement? for some implementation, applying the conditions on every exporting clusters are not feasible, and not necessary. The conflicted service from other clusters should not stop/impact my existing exporting traffic.

zhiying-lin: could you please share the conversation link with me? i missed that

tpantelis: I assume the meeting was recorded but I don't know where to access them.

zhiying-lin: i see, the recent recording was not uploaded, i;ll bring up in today's APAC sig meeting, thank you for the info!

PR #112, opened 2025-06-27
apis: add ServiceExport condition type/reasons

MrFreezeex:
Proposal for ServiceExport conditon type/reasons

suggested by @mikemorris here https://github.com/kubernetes/enhancements/pull/5438#issuecomment-3009121169 also this initial proposal is mainly based on GatewayAPI code adapted for our needs.

### DIFF
--- a/pkg/apis/v1alpha1/serviceexport.go
+++ b/pkg/apis/v1alpha1/serviceexport.go
@@ -67,7 +67,7 @@ const (
 	// This will be false if the service is found to be unexportable
 	// (ExternalName, not found).
 	//
-	// Deprecated: use ServiceExportConditionAccepted instead
+	// Deprecated: use ServiceExportConditionValid instead
 	ServiceExportValid = "Valid"
 	// ServiceExportConflict means that there is a conflict between two
 	// exports for the same Service. When "True", the condition message
@@ -76,7 +76,7 @@ const (
 	// Users should not expect detailed per-cluster information in the
 	// conflict message.
 	//
-	// Deprecated: use ServiceExportConditionConflicted instead
+	// Deprecated: use ServiceExportConditionConflict instead
 	ServiceExportConflict = "Conflict"
 )
 
@@ -114,14 +114,14 @@ func NewServiceExportCondition(t ServiceExportConditionType, status metav1.Condi
 }
 
 const (
-	// ServiceExportConditionAccepted is true when the Service Export is accepted.
+	// ServiceExportConditionValid is true when the Service Export is valid.
 	// This does not indicate whether or not the configuration has been exported
 	// to a control plane / data plane.
 	//
 	//
 	// Possible reasons for this condition to be true are:
 	//
-	// * "Accepted"
+	// * "Valid"
 	//
 	// Possible reasons for this condition to be False are:
 	//
@@ -131,33 +131,31 @@ const (
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
-	ServiceExportConditionAccepted ServiceExportConditionType = "Accepted"
+	ServiceExportConditionValid ServiceExportConditionType = "Valid"
 
-	// ServiceExportReasonAccepted is used with the "Accepted" condition when the
+	// ServiceExportReasonValid is used with the "Valid" condition when the
 	// condition is True.
-	ServiceExportReasonAccepted ServiceExportConditionReason = "Accepted"
+	ServiceExportReasonValid ServiceExportConditionReason = "Valid"
 
-	// ServiceExportReasonNoService is used with the "Accepted" condition when
+	// ServiceExportReasonNoService is used with the "Valid" condition when
 	// the associated Service does not exist.
 	ServiceExportReasonNoService ServiceExportConditionReason = "NoService"
 
-	// ServiceExportReasonInvalidServiceType is used with the "Accepted"
+	// ServiceExportReasonInvalidServiceType is used with the "Valid"
 	// condition when the associated Service has an invalid type
 	// (per the KEP at least the ExternalName type).
 	ServiceExportReasonInvalidServiceType ServiceExportConditionReason = "InvalidServiceType"
 )
 
 const (
-	// ServiceExportConditionExported is true when the service is exported to some
-	// control plane or data plane. This condition might not makes sense
-	// for every implementation, particularly if they function in a "pull mode"
-	// rather than actually exporting the service, and in that case implementations
-	// do not have do use this Condition type at all.
+	// ServiceExportConditionReady is true when the service is exported
+	// to some control plane or data plane or ready to be pulled.
 	//
 	//
 	// Possible reasons for this condition to be true are:
 	//
 	// * "Exported"
+	// * "Ready"
 	//
 	// Possible reasons for this condition to be False are:
 	//
@@ -171,24 +169,32 @@ const (
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
-	ServiceExportConditionExported ServiceExportConditionType = "Exported"
+	ServiceExportConditionReady ServiceExportConditionType = "Ready"
 
-	// ServiceExportReasonExported is used with the "Exported" condition
-	// when the condition is True.
+	// ServiceExportReasonExported is used with the "Ready" condition
+	// when the condition is True and the service has been exported.
+	// This would be used when an implementation exports a service
+	// to a control plane or data plane.
 	ServiceExportReasonExported ServiceExportConditionReason = "Exported"
 
-	// ServiceExportReasonPending is used with the "Exported" condition
+	// ServiceExportReasonReady is used with the "Ready" condition
+	// when the condition is True and the service has been exported.
+	// This would typically be used in an implementation that uses a
+	// pull model.
+	ServiceExportReasonReady ServiceExportConditionReason = "Ready"
+
+	// ServiceExportReasonPending is used with the "Ready" condition
 	// when the service is in the process of being exported.
 	ServiceExportReasonPending ServiceExportConditionReason = "Pending"
 
-	// ServiceExportReasonFailed is used with the "Exported" condition
+	// ServiceExportReasonFailed is used with the "Ready" condition
 	// when the service failed to be exported with the message providing
 	// the specific reason.
 	ServiceExportReasonFailed ServiceExportConditionReason = "Failed"
 )
 
 const (
-	// ServiceExportConditionConflicted indicates that some property of an
+	// ServiceExportConditionConflict indicates that some property of an
 	// exported service has conflicting values across the constituent
 	// ServiceExports. This condition must be at least raised on the
 	// conflicting ServiceExport and is recommended to be raised on all on
@@ -214,40 +220,40 @@ const (
 	// Controllers may raise this condition with other reasons,
 	// but should prefer to use the reasons listed above to improve
 	// interoperability.
-	ServiceExportConditionConflicted ServiceExportConditionType = "Conflicted"
+	ServiceExportConditionConflict ServiceExportConditionType = "Conflict"
 
-	// ServiceExportReasonPortConflict is used with the "Conflicted" condition
+	// ServiceExportReasonPortConflict is used with the "Conflict" condition
 	// when the exported service has a conflict related to port configuration.
 	// This includes when ports on resulting imported services would have
 	// duplicated names (including unnamed/empty name) or duplicated
 	// port/protocol pairs.
 	ServiceExportReasonPortConflict ServiceExportConditionReason = "PortConflict"
 
-	// ServiceExportReasonTypeConflict is used with the "Conflicted" condition
+	// ServiceExportReasonTypeConflict is used with the "Conflict" condition
 	// when the exported service has a conflict related to the service type
 	// (eg headless vs non-headless).
 	ServiceExportReasonTypeConflict ServiceExportConditionReason = "TypeConflict"
 
-	// ServiceExportReasonSessionAffinityConflict is used with the "Conflicted"
+	// ServiceExportReasonSessionAffinityConflict is used with the "Conflict"
 	// condition when the exported service has a conflict related to session affinity.
 	ServiceExportReasonSessionAffinityConflict ServiceExportConditionReason = "SessionAffinityConflict"
 
 	// ServiceExportReasonSessionAffinityConfigConflict is used with the
-	// "Conflicted" condition when the exported service has a conflict related
+	// "Conflict" condition when the exported service has a conflict related
 	// to session affinity config.
 	ServiceExportReasonSessionAffinityConfigConflict ServiceExportConditionReason = "SessionAffinityConfigConflict"
 
-	// ServiceExportReasonLabelsConflict is used with the "Conflicted"
+	// ServiceExportReasonLabelsConflict is used with the "Conflict"
 	// condition when the ServiceExport has a conflict related to exported
 	// labels.
 	ServiceExportReasonLabelsConflict ServiceExportConditionReason = "LabelsConflict"
 
-	// ServiceExportReasonAnnotationsConflict is used with the "Conflicted"
+	// ServiceExportReasonAnnotationsConflict is used with the "Conflict"
 	// condition when the ServiceExport has a conflict related to exported
 	// annotations.
 	ServiceExportReasonAnnotationsConflict ServiceExportConditionReason = "AnnotationsConflict"
 
-	// ServiceExportReasonNoConflicts is used with the "Conflicted" condition
+	// ServiceExportReasonNoConflicts is used with the "Conflict" condition
 	// when the condition is False.
 	ServiceExportReasonNoConflicts ServiceExportConditionReason = "NoConflicts"
 )
--- a/pkg/apis/v1alpha1/serviceexport.go
+++ b/pkg/apis/v1alpha1/serviceexport.go
@@ -66,6 +66,8 @@ const (
 	// service export has been recognized as valid by an mcs-controller.
 	// This will be false if the service is found to be unexportable
 	// (ExternalName, not found).
+	//
+	// Deprecated: use ServiceExportConditionAccepted instead
 	ServiceExportValid = "Valid"
 	// ServiceExportConflict means that there is a conflict between two
 	// exports for the same Service. When "True", the condition message
@@ -73,6 +75,8 @@ const (
 	// field(s) under contention, which cluster won, and why.
 	// Users should not expect detailed per-cluster information in the
 	// conflict message.
+	//
+	// Deprecated: use ServiceExportConditionConflicted instead
 	ServiceExportConflict = "Conflict"
 )
 
@@ -88,3 +92,162 @@ type ServiceExportList struct {
 	// +listType=set
 	Items []ServiceExport `json:"items"`
 }
+
+// ServiceExportConditionType is a type of condition associated with a
+// ServiceExport. This type should be used with the ServiceExportStatus.Conditions
+// field.
+type ServiceExportConditionType string
+
+// ServiceExportConditionReason defines the set of reasons that explain why a
+// particular ServiceExport condition type has been raised.
+type ServiceExportConditionReason string
+
+// NewServiceExportCondition creates a new ServiceExport condition
+func NewServiceExportCondition(t ServiceExportConditionType, status metav1.ConditionStatus, reason ServiceExportConditionReason, msg string) metav1.Condition {
+	return metav1.Condition{
+		Type:               string(t),
+		Status:             status,
+		Reason:             string(reason),
+		Message:            msg,
+		LastTransitionTime: metav1.Now(),
+	}
+}
+
+const (
+	// ServiceExportConditionAccepted is true when the Service Export is accepted.
+	// This does not indicate whether or not the configuration has been exported
+	// to a control plane / data plane.
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Accepted"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NoService"
+	// * "InvalidServiceType"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionAccepted ServiceExportConditionType = "Accepted"
+
+	// ServiceExportReasonAccepted is used with the "Accepted" condition when the
+	// condition is True.
+	ServiceExportReasonAccepted ServiceExportConditionReason = "Accepted"
+
+	// ServiceExportReasonNoService is used with the "Accepted" condition when
+	// the associated Service does not exist.
+	ServiceExportReasonNoService ServiceExportConditionReason = "NoService"
+
+	// ServiceExportReasonInvalidServiceType is used with the "Accepted"
+	// condition when the associated Service has an invalid type
+	// (per the KEP at least the ExternalName type).
+	ServiceExportReasonInvalidServiceType ServiceExportConditionReason = "InvalidServiceType"
+)
+
+const (
+	// ServiceExportConditionExported is true when the service is exported to some
+	// control plane or data plane. This condition might not makes sense
+	// for every implementation, particularly if they function in a "pull mode"
+	// rather than actually exporting the service, and in that case implementations
+	// do not have do use this Condition type at all.
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Exported"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Pending"
+	// * "Failed"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Pending"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionExported ServiceExportConditionType = "Exported"
+
+	// ServiceExportReasonExported is used with the "Exported" condition
+	// when the condition is True.
+	ServiceExportReasonExported ServiceExportConditionReason = "Exported"
+
+	// ServiceExportReasonPending is used with the "Exported" condition
+	// when the service is in the process of being exported.
+	ServiceExportReasonPending ServiceExportConditionReason = "Pending"
+
+	// ServiceExportReasonFailed is used with the "Exported" condition
+	// when the service failed to be exported with the message providing
+	// the specific reason.
+	ServiceExportReasonFailed ServiceExportConditionReason = "Failed"
+)
+
+const (
+	// ServiceExportConditionConflicted indicates that some property of an
+	// exported service has conflicting values across the constituent
+	// ServiceExports. This condition must be at least raised on the
+	// conflicting ServiceExport and is recommended to be raised on all on
+	// all the constituent ServiceExports if feasible.
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "PortConflict"
+	// * "TypeConflict"
+	// * "SessionAffinityConflict"
+	// * "SessionAffinityConfigConflict"
+	// * "AnnotationsConflict"
+	// * "LabelsConflict"
+	//
+	// When multiple conflicts occurs the above reasons may be combined
+	// using commas.
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflicted ServiceExportConditionType = "Conflicted"
+
+	// ServiceExportReasonPortConflict is used with the "Conflicted" condition
+	// when the exported service has a conflict related to port configuration.
+	// This includes when ports on resulting imported services would have
+	// duplicated names (including unnamed/empty name) or duplicated
+	// port/protocol pairs.
+	ServiceExportReasonPortConflict ServiceExportConditionReason = "PortConflict"
+
+	// ServiceExportReasonTypeConflict is used with the "Conflicted" condition
+	// when the exported service has a conflict related to the service type
+	// (eg headless vs non-headless).
+	ServiceExportReasonTypeConflict ServiceExportConditionReason = "TypeConflict"
+
+	// ServiceExportReasonSessionAffinityConflict is used with the "Conflicted"
+	// condition when the exported service has a conflict related to session affinity.
+	ServiceExportReasonSessionAffinityConflict ServiceExportConditionReason = "SessionAffinityConflict"
+
+	// ServiceExportReasonSessionAffinityConfigConflict is used with the
+	// "Conflicted" condition when the exported service has a conflict related
+	// to session affinity config.
+	ServiceExportReasonSessionAffinityConfigConflict ServiceExportConditionReason = "SessionAffinityConfigConflict"
+
+	// ServiceExportReasonLabelsConflict is used with the "Conflicted"
+	// condition when the ServiceExport has a conflict related to exported
+	// labels.
+	ServiceExportReasonLabelsConflict ServiceExportConditionReason = "LabelsConflict"
+
+	// ServiceExportReasonAnnotationsConflict is used with the "Conflicted"
+	// condition when the ServiceExport has a conflict related to exported
+	// annotations.
+	ServiceExportReasonAnnotationsConflict ServiceExportConditionReason = "AnnotationsConflict"
+
+	// ServiceExportReasonNoConflicts is used with the "Conflicted" condition
+	// when the condition is False.
+	ServiceExportReasonNoConflicts ServiceExportConditionReason = "NoConflicts"
+)


pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
 }
+
+// ServiceExportConditionType is a type of condition associated with ServiceExport.
+// This type should be used with the ServiceExportStatus.Conditions
+// field.
+type ServiceExportConditionType string
+
+// ServiceExportConditionReason defines the set of reasons that explain
+// why a particular conflict condition type has been raised.
+type ServiceExportConditionReason string
```
tpantelis, line 103: The types don't really add any value b/c the Condition fields are still string and it forces a type conversion to string. In fact, we used to have `ServiceExportConditionType` when we had `ServiceExportCondition` but it was [removed](https://github.com/kubernetes-sigs/mcs-api/pull/33) when we switched to `metav1.Condition`.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// interoperability.
+	ServiceExportConditionExported ServiceExportConditionType = "Exported"
+
+	// ServiceExportReasonExported is used with the "Exported" condition
+	// when the condition is True.
+	ServiceExportReasonExported ServiceExportConditionReason = "Exported"
+
+	// ServiceExportReasonInvalid is used with the "Exported" condition
+	// when the exported service is invalid and will not be exported.
+	ServiceExportReasonInvalid ServiceExportConditionReason = "Invalid"
```
tpantelis, line 168: This is reflected in the `Valid` condition. If it's not valid then you wouldn't even try to export it in which case you wouldn't need to add the `Exported` condition at all.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflict ServiceExportConditionType = "Conflict"
+
+	// ServiceExportReasonPortsConflict is used with the "Conflict" condition
+	// when the exported service have a conflict related to ports.
+	ServiceExportReasonPortsConflict ServiceExportConditionReason = "PortsConflict"
```
mikemorris, line 201: Adding the specific port conflict language here from KEP-1645 would be helpful in determining when this condition should be set.

```suggestion
	// ServiceExportReasonPortConflict is used with the "Conflicted" condition
	// when a derived service would have a conflicting port configuration. This
	// includes when ports on exported services have duplicated names (including
	// unnamed/empty name) for different protocol and port number pairs. The
	// conflict resolution policy will determine which of the duplicated ports are
	// used by the ServiceImport.
	ServiceExportReasonPortConflict ServiceExportConditionReason = "PortConflict"
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "Exported"
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Invalid"
+	// * "Pending"
+	//
```
tpantelis, line 152: I think we need a `Failed` reason as well to reflect an error that prevented export (eg API server not available).

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionExported ServiceExportConditionType = "Exported"
+
+	// ServiceExportReasonExported is used with the "Exported" condition
+	// when the condition is True.
+	ServiceExportReasonExported ServiceExportConditionReason = "Exported"
```
tpantelis, line 164: How about `Success` so we don't duplicate the type name?

pkg/apis/v1alpha1/serviceexport.go
mikemorris: Reading the KEP more carefully it makes sense to me that the `union` language and conflict resolution policy was intended to prevent disrupted traffic in the case of non-matching exported services, but we likely want some conformance testing around if/how unioned ports from non-matching backends should be routed (should `foo.clusterset.local:80` potentially fail if port 80 is only exposed on some subset of backends and a backend without that port open is selected, or should an implementation ensure requests to that port are always routed to eligible backends?)

Defining that behavior more precisely shouldn't block this change though.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// be raised on all on all the constituent `ServiceExport`s if feasible.
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "PortsConflict"
+	// * "TypeConflict"
+	// * "SessionAffinityConflict"
+	// * "SessionAffinityConfigConflict"
+	// * "MultipleConflicts"
```
mikemorris, line 188:
```suggestion
	// * "Conflicted"
	// * "PortConflict"
	// * "TypeConflict"
	// * "SessionAffinityConflict"
	// * "SessionAffinityConfigConflict"
```

I'd suggest `Conflicted` (for parity with the similar Gateway API [ListenerConditionType](https://github.com/kubernetes-sigs/gateway-api/blob/e1310bbd66c54b757d28ee65cf363825f6189ca5/apisx/v1alpha1/xlistenerset_types.go#L415) and tense partity with `Exported` on this resource) as a general-purpose reason which could be used when multiple conflicts are present or when an implementation-specific conflict not covered by the standard reasons may be present (with details provided in the `message` field).

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflict ServiceExportConditionType = "Conflict"
```
mikemorris, line 197:
```suggestion
	ServiceExportConditionConflicted ServiceExportConditionType = "Conflicted"
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -66,13 +66,17 @@ const (
...
+	// Deprecated: use ServiceExportConditionValid instead
 	ServiceExportValid = "Valid"
 	// ServiceExportConflict means that there is a conflict between two
 	// exports for the same Service. When "True", the condition message
 	// should contain enough information to diagnose the conflict:
 	// field(s) under contention, which cluster won, and why.
 	// Users should not expect detailed per-cluster information in the
 	// conflict message.
+	//
+	// Deprecated: use ServiceExportConditionConflict instead
```
tpantelis, line 79: I think the existing constant names are fine but if the consensus is to rename them then just change the existing constants instead of deprecating. I think deprecating is overkill in this case. It's not a big deal to modify a constant name after bumping the mcs-api version. Also, if one has linting that flags the use of deprecated fields then it would have to be changed anyway.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+
+	// ServiceExportReasonSessionAffinityConfigConflict is used with the
+	// "Conflict" condition when the exported service have a conflict related
+	// to session affinity config.
+	ServiceExportReasonSessionAffinityConfigConflict ServiceExportConditionReason = "SessionAffinityConfigConflict"
+
+	// ServiceExportReasonMultipleConflicts is used with the "Conflict"
+	// condition when the exported service have multiple conflicts if the
+	// implementation coalesce those.
+	ServiceExportReasonMultipleConflicts ServiceExportConditionReason = "MultipleConflicts"
```
mikemorris, line 219:
```suggestion
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflict ServiceExportConditionType = "Conflict"
+
```
mikemorris, line 198:
```suggestion

	// ServiceExportReasonConflicted is used with the "Conflicted"
	// condition when the exported service has multiple conflicts if the
	// implementation coalesces those or if an implementation-specific
	// conflict not matching one of the standard reasons is present.
	ServiceExportReasonConflicted ServiceExportConditionReason = "Conflicted"
	
```

mikemorris (commented):
Minor suggestions, overall looks like a great improvement!

pkg/apis/v1alpha1/serviceexport.go
mikemorris: Duplicating the type name is actually quite typical for the similar status conditions in Gateway API.

pkg/apis/v1alpha1/serviceexport.go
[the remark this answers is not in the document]
MrFreezeex: While I don't mind adding that as we won't probably do that in Cilium (at least right now) I don't think the conditions lib support has great support for that

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Invalid"
+	// * "Pending"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Pending"
+	//
```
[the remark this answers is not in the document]
tpantelis: The name `Exported` is fine too. But this condition type seems implementation-specific so perhaps we shouldn't define it here.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// least raised on the conflicting ServiceExport and is recommended to
+	// be raised on all on all the constituent `ServiceExport`s if feasible.
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "PortsConflict"
+	// * "TypeConflict"
+	// * "SessionAffinityConflict"
+	// * "SessionAffinityConfigConflict"
```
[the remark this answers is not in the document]
mikemorris: The current proposal is aligned with naming conventions for similar condition reasons in Gateway API https://github.com/kubernetes-sigs/gateway-api/blob/e1310bbd66c54b757d28ee65cf363825f6189ca5/apis/v1/gateway_types.go#L1168-L1183

pkg/apis/v1alpha1/serviceexport.go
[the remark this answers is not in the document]
mikemorris: Yea makes sense to me, `Valid`/`Accepted` is likely the most relevant to check in conformance.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: I'm wondering if `Pending` wouldn't work in this case? Also not sure if this should be `Failing` or `Failed`.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Well Gateway-API does that and indeed implementations need a cast to string everywhere, I don't know if we want to actually do that or not

pkg/apis/v1alpha1/serviceexport.go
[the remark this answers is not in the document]
tpantelis: Not sure I follow what you mean... the `Reason` field allows comma.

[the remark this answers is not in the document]
MrFreezeex: Ah yep nevermind I thought that it was possible to search for a specific reason with some conditions related functions but actually there is nothing like that

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: I applied @tpantelis suggestion to add commas let me know if you want to maintain this

pkg/apis/v1alpha1/serviceexport.go
tpantelis: I don't think so. `Pending` means we haven't tried it yet.

pkg/apis/v1alpha1/serviceexport.go
tpantelis: Perhaps but the way Gateway API does things doesn't necessarily mean it's the best way ... 😄

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,133 @@ type ServiceExportList struct {
...
+	// * "InvalidServiceType"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionValid ServiceExportConditionType = "Valid"
+
+	// ServiceExportReasonValid is used with the "Valid" condition when the
+	// condition is True.
+	ServiceExportReasonValid ServiceExportConditionReason = "Valid"
```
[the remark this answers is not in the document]
tpantelis: `Accepted` makes sense as well.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: To me the removal of those should probably be in a new version of the apis (v1alpha2 or v1beta1) to try to be mindful of people importing the package and using those. If you have linting that flags deprecated I would tend to say that it actually is achieving what's expected: aka you should probably change your conditions usage :thinking:.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Sounds good, I will add that!

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: I think that would probably makes sense in a context where you have some internal processing in the implementations and want to type your function rather than returning a plain string.

Although we don't do that in Cilium for instance so for us too the value of keeping that would be pretty much none (while requiring to cast in many places). I am guessing it would be pretty much the same for you @tpantelis?

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: My 2 cents is that our usage of conditions was probably not that great so far (at least for Cilium) and defaulting to GatewayAPI things except if we find good reasons to not do it on a case by case basis. And in that case both options reasonable enough :man_shrugging:.

pkg/apis/v1alpha1/serviceexport.go
tpantelis: Yes - the type casting became too much. Now we could provide a helper function like NewServiceExportValidCondition(reason ServiceExportConditionReason …) to hide the type casts.  But, in the end, any custom string not defined here is still valid so the type doesn’t really add value.

pkg/apis/v1alpha1/serviceexport.go
mikemorris: Might still be useful as a generic case, but no strong feelings.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Ok then I removed those, we could still reconsider this if we find some other benefits

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Well for instance there's a bunch of deprecation here related to conditions: https://github.com/kubernetes-sigs/gateway-api/blob/e1310bbd66c54b757d28ee65cf363825f6189ca5/apis/v1/gateway_types.go

My reasoning is that when project would use a newer version of the CRD they will import some other go package meaning that they will already need to actively change some code anyway and thus we would be able to not carry over certain deprecation such as this one.

For small version bump though I don't think we should make users of this library change their code by dropping those const IMO.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Does a project that have some very specific conflict should be using this generic reason or create their own though? If creating their own I would say that this shouldn't be necessary right?

pkg/apis/v1alpha1/serviceexport.go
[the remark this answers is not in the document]
tpantelis: Yes. I think the fact that it allows a comma is an indication that concatenation is deemed OK (not sure why else you would use a comma in an otherwise CamelCase string). It also allows colon.

pkg/apis/v1alpha1/serviceexport.go
mikemorris: > I think that would probably makes sense in a context where you have some internal processing in the implementations and want to type your function rather than returning a plain string.

Yea, precisely this and avoiding accidentally returning a condition of the wrong type (e.g. a ServiceImport reason for a ServiceExport type). In practice I think the typecasts aren't particularly annoying (and I believe are elided in compilation), here's an [example](https://github.com/istio/istio/blob/34895c89f0e44758da6f6d22d8b8f56754fd4f31/pilot/pkg/config/kube/gateway/conditions.go#L147-L156) from Istio's Gateway API implementation.

I don't think they're essential though, just a minor detail to guide implementers of the spec.

tpantelis: Yeah the types can make sense when used in functions. Looking at the Submariner code, we do actually have a function `newServiceExportCondition(condType string, status metav1.ConditionStatus, reason, msg string) metav1.Condition` so we could simply change the signature. It's the unit tests which have more type casting.

If we want to define types then I think it would be helpful to provide a `NewServiceExportCondition(condType ServiceExportConditionType, status metav1.ConditionStatus, reason ServiceExportConditionReason, msg string) metav1.Condition` function that uses the types to further guide implementors.

tpantelis:
> Yep good point that seems reasonable that as long (or at least for some time) as we support v1alpha1 ...

yeah which is why I think it makes sense to change the condition type names in **v1alpha2** along with the other proposed (potentially breaking) `ServiceImport` changes.

MrFreezeex:
> > Yep good point that seems reasonable that as long (or at least for some time) as we support v1alpha1 ...
> 
> yeah which is why I think it makes sense to change the condition type names in **v1alpha2** along with the other proposed (potentially breaking) `ServiceImport` changes.

The thing is that those future `ServiceImport` breaking changes are actually not breaking if we actually do this "mirror fields" approach. That being said you are right that it should be treated similarly indeed, so we should decide in what API those `ServiceImport` "breaking changes" would go first.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	//
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflicted = "Conflicted"
+
+	// ServiceExportReasonPortConflict is used with the "Conflicted" condition
+	// when the exported service have a conflict related to port configuration.
```
tpantelis, line 195:
```suggestion
	// when the exported service has a conflict related to port configuration.
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	// This includes when ports on resulting imported services would have
+	// duplicated names (including unnamed/empty name) or duplicated
+	// port/protocol pairs.
+	ServiceExportReasonPortConflict = "PortConflict"
+
+	// ServiceExportReasonTypeConflict is used with the "Conflicted" condition
+	// when the exported service have a conflict related to type.
```
tpantelis, line 202:
```suggestion
	// when the exported service has a conflict related to the service type (eg headless vs non-headless).
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	ServiceExportReasonTypeConflict = "TypeConflict"
+
+	// ServiceExportReasonSessionAffinityConflict is used with the "Conflicted"
+	// condition when the exported service have a conflict related to session affinity.
```
tpantelis, line 206:
```suggestion
	// condition when the exported service has a conflict related to session affinity.
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	ServiceExportReasonSessionAffinityConflict = "SessionAffinityConflict"
+
+	// ServiceExportReasonSessionAffinityConfigConflict is used with the
+	// "Conflicted" condition when the exported service have a conflict related
```
tpantelis, line 210:
```suggestion
	// "Conflicted" condition when the exported service has a conflict related
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	// to session affinity config.
+	ServiceExportReasonSessionAffinityConfigConflict = "SessionAffinityConfigConflict"
+
+	// ServiceExportReasonLabelsConflict is used with the "Conflicted"
+	// condition when the exported service have a conflict related to labels.
```
tpantelis, line 215:
```suggestion
	// condition when the exported service has a conflict related to ServiceExport labels.
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	ServiceExportReasonLabelsConflict = "LabelsConflict"
+
+	// ServiceExportReasonAnnotationsConflict is used with the "Conflicted"
+	// condition when the exported service have a conflict related to annotations.
```
tpantelis, line 219:
```suggestion
	// condition when the exported service has a conflict related to ServiceExport annotations.
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	ServiceExportReasonSessionAffinityConflict = "SessionAffinityConflict"
+
+	// ServiceExportReasonSessionAffinityConfigConflict is used with the
+	// "Conflicted" condition when the exported service have a conflict related
+	// to session affinity config.
+	ServiceExportReasonSessionAffinityConfigConflict = "SessionAffinityConfigConflict"
+
+	// ServiceExportReasonLabelsConflict is used with the "Conflicted"
+	// condition when the exported service have a conflict related to labels.
+	ServiceExportReasonLabelsConflict = "LabelsConflict"
```
tpantelis, line 216: I think this should be treated similar as `PortConflict`.

```suggestion
	ServiceExportReasonLabelConflict = "LabelConflict"
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+
+	// ServiceExportReasonAnnotationsConflict is used with the "Conflicted"
+	// condition when the exported service have a conflict related to annotations.
+	ServiceExportReasonAnnotationsConflict = "AnnotationsConflict"
```
tpantelis, line 220:
```suggestion
	ServiceExportReasonAnnotationConflict = "AnnotationConflict"
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,134 @@ type ServiceExportList struct {
...
+	// ServiceExportReasonFailed is used with the "Exported" condition
+	// when the service failed to be exported.
+	ServiceExportReasonFailed = "Failed"
+)
+
+const (
+	// ServiceExportConditionConflicted indicates that the controller was unable
+	// to resolve conflict for a ServiceExport. This condition must be at
+	// least raised on the conflicting ServiceExport and is recommended to
+	// be raised on all on all the constituent `ServiceExport`s if feasible.
```
tpantelis, line 170: It indicates there was some conflicting aspect between 2 or more exported services but it doesn't mean it wasn't resolved. Also I'm not sure the second statement re: where the condition should be raised is really relevant here.

```suggestion
	// ServiceExportConditionConflicted indicates that some property of an exported service has conflicting 
        //  values across the constituent ServiceExports.
```

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Added as suggested!

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,158 @@ type ServiceExportList struct {
...
+type ServiceExportConditionType string
+
+// ServiceExportConditionReason defines the set of reasons that explain why a
+// particular ServiceExport condition type has been raised.
+type ServiceExportConditionReason string
+
+// NewServiceExportCondition creates a new ServiceExport condition
+func NewServiceExportCondition(t ServiceExportConditionType, status metav1.ConditionStatus, reason ServiceExportConditionReason, msg string) metav1.Condition {
+	return metav1.Condition{
+		Type:    string(t),
```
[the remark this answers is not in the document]
mikemorris: I guess it doesn't hurt because passing a condition with LastTransitionTime set to that function will be ignored if the status field is unchanged, but I would lean toward leaving this out and deferring to handling setting this accurately in implementation controller logic.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,159 @@ type ServiceExportList struct {
...
+	// interoperability.
+	ServiceExportConditionAccepted ServiceExportConditionType = "Accepted"
+
+	// ServiceExportReasonAccepted is used with the "Accepted" condition when the
+	// condition is True.
+	ServiceExportReasonAccepted ServiceExportConditionReason = "Accepted"
+
+	// ServiceExportReasonNoService is used with the "Accepted" condition when
+	// the associated Service does not exist.
+	ServiceExportReasonNoService ServiceExportConditionReason = "NoService"
```
mikemorris, line 142: Is there potentially any RBAC reason that (controller can't read Service resources for some reason) that could cause this?
```suggestion
	// ServiceExportReasonNoService is used with the "Accepted" condition when
	// the Service to be exported could not be found.
	ServiceExportReasonServiceNotFound ServiceExportConditionReason = "ServiceNotFound"
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,159 @@ type ServiceExportList struct {
...
+	// using commas.
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "NoConflicts"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionConflicted ServiceExportConditionType = "Conflicted"
```
zhiying-lin, line 214: ditto, not obvious benefits to change to "conflicted" compared with "Conflict"

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,159 @@ type ServiceExportList struct {
...
+	ServiceExportReasonNoService ServiceExportConditionReason = "NoService"
+
+	// ServiceExportReasonInvalidServiceType is used with the "Accepted"
+	// condition when the associated Service has an invalid type
+	// (per the KEP at least the ExternalName type).
+	ServiceExportReasonInvalidServiceType ServiceExportConditionReason = "InvalidServiceType"
+)
+
+const (
+	// ServiceExportConditionExported is true when the service is exported to some
```
zhiying-lin, line 151: Before in the KEP, we treat the "Valid" true and "Conflict" false as "exported", which means it's visible outside of the local cluster.

What "Pending" and "Failed" mean may be related to "Import"? 
Does it mean that it can be exported to every imported cluster? which it's a dynamic state and very hard to track.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,159 @@ type ServiceExportList struct {
...
+	//
+	//
+	// Possible reasons for this condition to be true are:
+	//
+	// * "PortConflict"
+	// * "TypeConflict"
+	// * "SessionAffinityConflict"
+	// * "SessionAffinityConfigConflict"
+	// * "AnnotationsConflict"
+	// * "LabelsConflict"
```
zhiying-lin, line 202: hmm, it's not easy to extend, later we may support other field conflicts.

How about just defining the "ConflictFound" and putting the details in the message?

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,159 @@ type ServiceExportList struct {
...
+	//
+	// When multiple conflicts occurs the above reasons may be combined
+	// using commas.
```
zhiying-lin, line 205: I may be wrong, i did not see this pattern before. Usually we treat the reason as a string, not a string array and there's no agreed statement/util about to parse this string.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: I would prefer to have specific reasons IMO, you are free to use your own reasons in case your use case is not supported as stated in this file.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -66,13 +66,17 @@ const (
 	// service export has been recognized as valid by an mcs-controller.
 	// This will be false if the service is found to be unexportable
 	// (ExternalName, not found).
+	//
+	// Deprecated: use ServiceExportConditionAccepted instead
```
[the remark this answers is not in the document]
MrFreezeex: Do you actually read back the conditions to do something? If not I would expect very little migration effort needed here actually...

The conditions defined here predates Conditions being a Kubernetes type so it's a complete overhaul. We are also aligning ourselves with GatewayAPI works which use "Accepted" rather than "Valid"

pkg/apis/v1alpha1/serviceexport.go
mikemorris: >  Does it mean that it can be exported to every imported cluster? which it's a dynamic state and very hard to track.

Agreed that this semantics/state is likely impractical to express through the `Exported` condition - whether each ServiceImport is successfully created should likely be messaged through a `Ready` condition on each ServiceImport as proposed in https://github.com/kubernetes-sigs/mcs-api/pull/113

pkg/apis/v1alpha1/serviceexport.go
mikemorris: For parity with Gateway API and the [SIG-Architecture API conventions for status conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), which I believe did not exist yet when these early conditions were initially added, which explicitly recommend past-tense verbs.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: > I'm not entirely clear on if decentralized implementations (Cilium, Submariner) will have a use for this reason.

Cilium nop as we don't really export anything technically. Submariner I am not sure, I think I grasped from @tpantelis comment that yes but I let him confirm.

pkg/apis/v1alpha1/serviceexport.go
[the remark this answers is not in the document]
zhiying-lin: It definitely will involve a lot of provider code changes and it will break all the tests.

Since we're still in v1alpha1, it should be our last chance to make it right.

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: The conformance test generally present targeted conflict with one property conflicting so it should be relatively fine to make this into the conformance test. If there are test with multiple conflict we would probably not test the exact reason I would say. Also if you have particular situation that are specific to your implementation it would probably not be in the conformance test to begin with.

pkg/apis/v1alpha1/serviceexport.go
tpantelis: Submariner would use this condition and, in fact, already does except we currently name it `Ready`.

MrFreezeex: Actually maybe we should also consider using "Ready" rather than exported here?

It would probably be very similar to the current thing but have a very generic "Ready" reason which can be true in addition to the exported one. That way Cilium could use that wsith the "Ready" reason and things that have some generic logic about conditions (for instance: https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus) could know about the resource condition with the Ready type. What do you all think about this idea?

tpantelis: I’m fine with either one.

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,167 @@ type ServiceExportList struct {
...
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceExportConditionReady ServiceExportConditionType = "Ready"
+
+	// ServiceExportReasonExported is used with the "Ready" condition
+	// when the condition is True and the service are exported. This should be
+	// used when an implementation have to export a service to a control plane
+	// or data plane.
```
tpantelis, line 177:
```suggestion
	// when the condition is True and the service has been exported. This would be
	// used when an implementation exports a service to a control plane
	// or data plane.
```

pkg/apis/v1alpha1/serviceexport.go
```diff
@@ -88,3 +92,167 @@ type ServiceExportList struct {
...
+	ServiceExportReasonExported ServiceExportConditionReason = "Exported"
+
+	// ServiceExportReasonReady is used with the "Ready" condition
+	// when the condition is True and the service are exported. This would
+	// typically be used in an implementation that using a pull model.
```
tpantelis, line 182:
```suggestion
	// when the condition is True and the service has been exported. This would
	// typically be used in an implementation that uses a pull model.
```

pkg/apis/v1alpha1/serviceexport.go
MrFreezeex: Thanks!

lauralorenz:
Triage notes:
* The plan has been done YAY
* Ready for review again

PR #113, opened 2025-06-27
apis: add ServiceImport condition

MrFreezeex:
TODO:
- [x]  get consensus on ServiceImport condition / KEP proposing ServiceImport condition is merged (https://github.com/kubernetes/enhancements/pull/5439)
- [x] get consensus on the type/reasons proposed here

### DIFF
--- a/pkg/apis/v1alpha1/serviceimport.go
+++ b/pkg/apis/v1alpha1/serviceimport.go
@@ -120,6 +120,12 @@ type ServiceImportStatus struct {
 	// +listType=map
 	// +listMapKey=cluster
 	Clusters []ClusterStatus `json:"clusters,omitempty"`
+	// +optional
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
 // ClusterStatus contains service configuration mapped to a specific source cluster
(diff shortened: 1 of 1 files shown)


pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -141,3 +147,46 @@ type ServiceImportList struct {
...
+	ServiceImportConditionReady = "Ready"
+
+	// ServiceImportReasonReady is used with the "Ready" condition when the
+	// condition is True.
+	ServiceImportReasonReady = "Ready"
+
+	// ServiceExportReasonPartialIPFamilyCompatibility is used with the "Ready"
+	// condition when the local cluster partially supports the imported service
+	// IP families.
+	ServiceExportReasonPartialIPFamilyCompatibility = "PartialIPFamilyCompatibility"
```
mikemorris, line 182: Should this be set when the importing cluster offers _more_ IP families than are available for endpoints (as a warning that in a dual-stack cluster, an IPv6 only client may fail to connect to a ServiceImport with only IPv4 backends available)?

I think ^ makes more sense than the inverse (endpoints are available for IPFamilies other than those supported by the importing cluster, but at least some are available for a matching IPFamily (because in this case, all clients can route to a backend and the ServiceImport could for all practical purposes be considered `Ready`)

pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -141,3 +147,46 @@ type ServiceImportList struct {
...
+
+	// ServiceExportReasonProgressing is used with the "Ready" condition when
+	// the ServiceImport is in the process of being created or updated.
+	ServiceExportReasonProgressing = "Progressing"
```
mikemorris, line 186:
```suggestion
	// ServiceExportReasonPending is used with the "Ready" condition when
	// the ServiceImport is in the process of being created or updated.
	ServiceExportReasonPending = "Pending"
```

pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -141,3 +147,46 @@ type ServiceImportList struct {
...
+	//
+	// Possible reasons for this condition to be False are:
+	//
+	// * "Progressing"
+	// * "PartialIPFamilyCompatibility"
+	// * "IPFamilyNotSupported"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Progressing"
```
mikemorris, line 168:
```suggestion
	// * "Pending"
```

mikemorris (commented):
I think this makes a lot of sense! Minor notes but mostly LGTM.

pkg/apis/v1alpha1/serviceimport.go
MrFreezeex: This was the for the inverse but I wasn't that convinced either when I wrote that so you writing about that it should be rather `Ready` comfort me into dropping this actually :sweat_smile:.

> an IPv6 only client may fail to connect to a ServiceImport with only IPv4 backends available)?

I don't think that this can happen because it's only the Service IPFamily that you can select in a dual stack cluster, IIRC all the pods are pretty much required to be dual stack.

pkg/apis/v1alpha1/serviceimport.go
```diff
@@ -141,3 +147,59 @@ type ServiceImportList struct {
...
+	// * "IPFamilyNotSupported"
+	//
+	// Possible reasons for this condition to be Unknown are:
+	//
+	// * "Pending"
+	//
+	// Controllers may raise this condition with other reasons,
+	// but should prefer to use the reasons listed above to improve
+	// interoperability.
+	ServiceImportConditionReady ServiceImportType = "Ready"
```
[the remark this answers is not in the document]
MrFreezeex: Ah yep! it actually compile because `ServiceImportType` also does exist (it's for the type of the ServiceImport aka `ClusterSetIP` or `Headless`)

PR #116, opened 2025-07-28
crd: add symbol with CRD version

MrFreezeex:
Add a CRD version label that must be updated each time we update the corresponding CRD to facilitate go install.

In cilium we are doing this for our internal CRDs with some logic that check this label so that we don't accidentally downgrade the version and it would be super useful if we can have directly in the mcs-api repo for MCS-API CRDs to facilitate that check.


youngnick:
Yes, the apiVersion field is a bit coarse-grained and only really tracks breaking changes. So having a schema version that tracks if compatible, additive changes have been made is very useful. That's why we include a bundle version in Gateway API CRDs.

MrFreezeex:
> I'd also recommend noting the procedure down in some release instructions somewhere - otherwise it's easy to forget.

Hmm well there's not really a properly defined release so far, but maybe this could be the same as the git tag which would make it an actual release? It would also mean that we would want to make a new release/git tag almost every-time we change the CRD. WDYT @skitt?

skitt:
Since we want to track changes, I think the version should be bumped whenever there’s a change — so the bump should be part of the PR making the change, not part of the release. This also means that the CRD version won’t be the same as the project tag, but that’s fine.

We’ll use a version triplet in the same style as Gateway API.

MrFreezeex:
I updated the PR to start from v0.1.0 (and also adding the v prefix). It's very similar to GW-API the only thing is that GW-API use the term "bundle-version" and this is proposing the term "crd-schema-version".

I feel like since it's not part of the release and that GW-API has different bundles (experimental) the term "crd-schema-version" we are in slighly different scenario than GW-API which is why I am proposing a different term.

But I don't feel strongly about it so I'm happy to change the term if this means landing faster this and/or that "bundle-version" (or something else) is better liked (cc @mikemorris).

PR #117, opened 2025-08-20
Increase artificial delay for 2 cluster conformance tests

tpantelis:
It delays a bit in between exporting the services so the timestamps differ to make conflict checking deterministic. However 100 ms is too short as creation timestamps have seconds granularity so increase the delay to at least 1 sec.


MrFreezeex (approved):
Thanks! Makes sense to me :D

/lgtm

PR #118, opened 2025-08-28
Log general steps in the conformance tests

tpantelis:
Specifically, when a service is deployed, exported and unexported to help with debugging when a test fails. Of particular use is the service namespace/name, cluster name and timestamps for correlation when inspecting pod logs.


MrFreezeex (approved):
That looks super nice for debugging, thanks!
/lgtm

PR #119, opened 2025-08-28
Adjust the service type conflict conformance test

tpantelis:
...to only verify a `Conflict` condition is applied on each cluster.

In Submariner, we use the type of the first observed `ServiceExport` for the derived multi-cluster service. So if the service is exported in another cluster and its actual timestamp happens to be older than the first, we don't change the derived service to avoid disruption. This is in accordance with the [KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#global-properties) which recently augmented the conflict resolution policy to allow for favoring maintaining service continuity and avoiding potentially disruptive changes.

So given that, I relaxed the test so it doesn't verify the service type.


tpantelis:
> Hmm I am not sure I understand this, the type shouldn't change dynamically in this test or am I missing something/not reading something correctly?

It's not that it changes - it's the initial resolution of the type. We export the first cluster, wait a second then export the second. So based on `ServiceExport` timestamps, the first cluster's type should win. But if going by first observed, it's theoretically non-deterministic, ie it depends on timing of notifications and processing in the API server and controllers. I've still seen the second cluster's type win even delaying a second in between.

MrFreezeex (approved):
Ah ok I see! Could there be a way to make this test wait for like the ServiceImport to be created before exporting in the second cluster then? This should make this not racy sounds ok to me to relax this test too though!

/lgtm

tpantelis:
> Ah ok I see! Could there be a way to make this test wait for like the ServiceImport to be created before exporting in the second cluster then?

yeah I thought about something like that too but that seems a bit kludgy plus I think the important aspect of this test is to verify the implementation recognizes the conflict and applies the conflict condition.

skitt:
/approve

Thanks!

PR #120, opened 2025-08-28
Use new ServiceExport Condition definitions in conformance tests

tpantelis:


tpantelis:
> > yeah thought about that too but the Ready condition is not required and the spec doesn't mention it.
> 
> Hmm unless maybe if Valid is false I would expect Ready to always be defined :( maybe it should be in the KEP first though indeed

yeah I think we should first define the `Ready` condition in the KEP before requiring compliance.

tpantelis:
@MrFreezeex I changed the "Valid" condition constant. I also added a check for `ServiceExportConditionValid == true` in one of the tests.

skitt:
/approve

Thanks!

PR #121, opened 2025-08-29
Relax criteria for the "Local service not impacted" E2E test

tpantelis:
This test has been failing intermittently for a while:

```
STEP: verifying a local, unexported service is reachable via cluster.local

[FAILED] Timed out after 10.092s.
  There is no failure as the matcher passed to Eventually succeeded on its most recent iteration
  In [It] at: ./home/prow/go/src/sigs.k8s.io/mcs-api/e2e/localserviceimpact_test.go:167
```

So the `Eventually` in `checkAllClustersReachable`:

```
Eventually(func(g Gomega) {
...
}, "10s").MustPassRepeatedly(50).Should(Succeed())
```

does succeed but sometimes does not do so 50 times within 10s. So we need to relax these criteria, either increase the timeout or decrease the number of times. I opted for the latter, ie 20 times.

### DIFF
--- a/e2e/localserviceimpact_test.go
+++ b/e2e/localserviceimpact_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


MrFreezeex (approved):
Thank!
/lgtm

PR #124, opened 2025-09-04
Add a workflow to label issues

skitt:
This ensures new and reopened issues are labeled "sig/multicluster", which makes them show up in our triage filter.

Labels could also be added using issue templates, but users can always choose to create an issue using a blank template and thereby skip the automatic labeling. Using a workflow ensures all issues are caught.

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,0 +1,19 @@
+---
+name: Label issues
+on:
+  issues:
+    types:
+      - opened
+      - reopened
+jobs:
+  label_issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - run: gh issue edit "$NUMBER" -add-label "$LABELS"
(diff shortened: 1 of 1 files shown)


RainbowMango:
> This ensures new and reopened issues are labeled "sig/multicluster", which makes them show up in our triage filter.

Where can I find the triage filter thing? 
I guess you are trying to filter issues across multiple repos.

RainbowMango:
Yeah, thanks. 

Not an objection, just my feelings and suggestions about this approach.

Running a GitHub Action job for every new issue just to add a label seems a little bit heavy. In addition, [the link](https://github.com/pulls?q=is%3Aopen+label%3Asig%2Fmulticluster+) performs a global GitHub search for the `sig/multicluster` label could return irrelevant results if other projects are also using the same label.

Maybe what we need is a board like [SIG-Instrumentation](https://github.com/orgs/kubernetes-sigs/projects/110/views/1).

Issue #125, opened 2025-09-08
Use upstream CoreDNS v1.12.2


kahirokunn:
## Background

Multicluster is now part of the CoreDNS Kubernetes plugin (v1.12.2). We only need Corefile and RBAC to enable `clusterset.local`.

## Plan

- Drop custom build
- Use `registry.k8s.io/coredns/coredns:v1.12.2`
- Update Corefile
- Install MCS API CRDs before rolling out CoreDNS.

### Corefile Example

https://github.com/coredns/coredns/blob/master/plugin/kubernetes/README.md#examples

```corefile
kubernetes cluster.local clusterset.local {
    multicluster clusterset.local
}
````

patch script:

```bash
kubectl --kubeconfig "${KUBECONFIG}" get configmap -n kube-system coredns -o yaml | \
  sed -E 's/^([[:space:]]*)kubernetes cluster\.local (.*)$/\1kubernetes cluster.local clusterset.local \2\n\1   multicluster clusterset.local/' | \
  kubectl --kubeconfig "${KUBECONFIG}" replace -f-
```

### RBAC Example

```yaml
rules:
- apiGroups: ["multicluster.x-k8s.io"]
  resources: ["serviceimports"]
  verbs: ["list", "watch"]
```

patch script:

```bash
kubectl patch clusterrole system:coredns --type=json --patch '[
  {
    "op": "add",
    "path": "/rules/-",
    "value": {
      "apiGroups": ["multicluster.x-k8s.io"],
      "resources": ["serviceimports"],
      "verbs": ["list","watch"]
    }
  }
]'
```

---

## Adoption Status

For reference, I checked Kubernetes distributions that use CoreDNS v1.12.2 or later; it’s already being adopted, and it looks likely to become commonplace.

| Product  | First version determined to "default 1.12.2+" | Link                                                                                                                                             |
| -------- | --------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| **k0s**  | v1.32.6+k0s.0                                 | [https://github.com/k0sproject/k0s/releases/tag/v1.32.6%2Bk0s.0](https://github.com/k0sproject/k0s/releases/tag/v1.32.6%2Bk0s.0)                 |
[...]

PR #126, opened 2025-09-08
Use upstream CoreDNS v1.12.2 with multicluster support

kahirokunn:
Fixes: #125 

## Summary

Switch CoreDNS to upstream `registry.k8s.io/coredns/coredns:v1.12.2`.

## Motivation

Multicluster is now part of the upstream CoreDNS Kubernetes plugin.

## References

- CoreDNS Kubernetes plugin docs:  
  https://github.com/coredns/coredns/blob/master/plugin/kubernetes/README.md#examples

## Local Test

CleanShot 2025-09-08 at 13 34 07@2x


config/crd/multicluster.x-k8s.io_serviceimports.yaml
```diff
@@ -182,3 +178,61 @@ spec:
                   x-kubernetes-list-map-keys:
                     - cluster
                   x-kubernetes-list-type: map
+                conditions:
```
MrFreezeex, line 181: Outch, thanks for spotting those! Could you do a separate PR for your first commit here?

scripts/c1.yaml
```diff
@@ -9,5 +9,4 @@ nodes:
     - |
       kind: ClusterConfiguration
       dns:
-        imageRepository: multicluster
-        imageTag: latest
+        imageTag: v1.12.2
```
MrFreezeex, line 12: Could you add a small todo or fixme comment to remove this after Kubernetes 1.35 (see https://github.com/kubernetes/kubernetes/pull/132288)?

MrFreezeex (commented):
Thanks a lot for doing this! I left a few comments inline

demo/demo.sh
```diff
@@ -44,8 +44,8 @@ trap cleanup EXIT
 tmux send -t "$c1_pane" "${k1} logs -f mcs-api-controller" Enter
 
 desc "Create our service in each cluster"
-run "${k1} apply -f ${script_dir}/yaml/dep1.yaml -f ${script_dir}/yaml/svc.yaml"
-run "${k2} apply -f ${script_dir}/yaml/dep2.yaml -f ${script_dir}/yaml/svc.yaml"
+run "${k1} apply -f ./yaml/dep1.yaml -f ./yaml/svc.yaml"
+run "${k2} apply -f ./yaml/dep2.yaml -f ./yaml/svc.yaml"
```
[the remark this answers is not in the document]
kahirokunn: Sorry about that — it turned out I was missing `scripts_dir` (I mistakenly wrote `script_dir`), which caused the error. I ended up experimenting with relative paths, but the correct key is `scripts_dir`. Since I want the script to work regardless of where it’s executed, I’ve actually restored the `scripts_dir` setting instead.

MrFreezeex:
> @MrFreezeex Hi! I tried Merged the "Add TODOs for DNS imageTag updates" changes into "Improve CoreDNS configuration". Does this look okay to you? 👀

:wave:, ah sorry I realized my message was not clear enough when I was talking about "first commit" I meant the first commit of your PR so this one "Adopt upstream CoreDNS v1.12.2 and remove custom build" currently. Basically my main concern is that the first commit introduce changes that you then fix afterwards instead of introducing the right change directly.

Also technically about this part of my last message

> and remove the changes to "demo/demo.sh" in your first commit to have a dedicated commit similar to your existing commit "Update demo script to use absolute paths"

It could be fine to squash "Update demo script to use absolute paths" in the first commit of your PR too if you don't want to fiddle too much with patching your first commit and fixing up conflict.

MrFreezeex:
Thanks again that lgtm towards the commit content/structured! But the e2e tests seems to fails unfortunately and it seems relevant? I am not sure why though :(. Are you able to reproduce that locally somehow?

kahirokunn:
@MrFreezeex 
This is probably the cause. 👀 
https://github.com/kubernetes-sigs/mcs-api/commit/130eae2049cad1795d8e82141ee1cccbfc666404

kahirokunn:
However, I don't know the cause... 😢 
It works fine locally.

kahirokunn:
Thanks for the thoughtful review! You’re right—the e2e failures were related. When updating the Corefile patching to follow the Cilium documentation (prerequisites), I accidentally omitted the --kubeconfig flag, so the change only applied to the current context instead of both clusters. I’ve fixed this by adding --kubeconfig on both the get and replace calls. Thanks again for catching this!

MrFreezeex (approved):
Thanks a lot for your two PRs! Your changes are super helpful! I was planning to do something similar too in some time but you beat me to it :D :pray:.

/lgtm

kahirokunn:
Thanks everyone for the thoughtful reviews and support throughout this PR 🙏
I’ve learned a lot from the feedback and really appreciate the collaboration. Excited to keep improving things together! 🙌

PR #132, opened 2025-11-25
apis: conformance: add more conflict condition on asymetrical traffic

MrFreezeex:
This is the "implementation" of the following KEP change: https://github.com/kubernetes/enhancements/pull/5706.

It standardize the IPFamilyConflict condition which implementation may optionally use and mandate that PortConflict should be used if all ports are not identical (while not changing the union behavior).


skitt:
/approve

Thanks!

PR #136, opened 2025-12-22
conformance: add connectity test for conflicting ports

### LINKED ISSUE #135: Conformance tests: optional connectivity tests for conflicting ports
Add optional connectivity test(s) corresponding to this part of the KEP in https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#service-port:
```
In that case, network traffic to a conflicting port should
only be directed to endpoints from constituent services that actually expose the port.
```

The two following tests might be interesting:
- cluster1 (oldest): svc ports: name:tcp port: 42 / cluster2 svc ports: name:tcp port: 4242 (same name but conflicting port) -> check that reaching port 42 should only get to cluster1
- cluster1 (oldest) svc ports: name:tcp port: 42 / cluster2 svc ports: name:tcp2 port:4242 (port union) -> check that reaching port 42 should only get to cluster1 and that reaching port 4242 only reach cluster2

MrFreezeex:
Extract the function to awaitServicePodIP and add two new connectivity tests with ports conflict with conflicting name and conflicting port number.

Fixes #135

### DIFF
--- a/conformance/connectivity.go
+++ b/conformance/connectivity.go
@@ -17,13 +17,12 @@ limitations under the License.
 package conformance
 
 import (
-	"context"
 	"fmt"
-	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 var _ = Describe("", func() {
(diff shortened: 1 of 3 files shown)


conformance/connectivity.go
```diff
@@ -90,24 +91,55 @@ var _ = Describe("", func() {
...
+			}
+		})
+	})
 
-				return servicePodIP
-			}, 20, 1).ShouldNot(BeEmpty(), "Service deployment pod was not allocated an IP")
+	Context("Connectivity to exported services with same port name but different port numbers on each cluster", func() {
+		BeforeEach(func() {
+			requireTwoClusters()
+			t.autoExportService = false
```
tpantelis: Minor thing but this is already set to false on L33 above.

MrFreezeex:
Thanks @tpantelis for all the suggestions, I applied everything and the code much cleaner now :saluting_face:. Let me know if there's anything else :eyes:

PR #143, opened 2026-03-11
Conformance: add IPv6 and dual-stack support

tpantelis:
Updates the conformance test suite to properly handle IPv4-only, IPv6-only, and dual-stack configurations. The tests adapt to the underlying IP family configuration of the K8s clusters.

Client-side changes:
- Iterate over each `ServiceImport` IP and service `IPFamily` to support multiple IP families in ClusterIP and headless DNS tests
- Use the IP family to determine the appropriate DNS record type (A for IPv4, AAAA for IPv6) in `nslookup` commands
- Add trailing dots to `nslookup` FQDNs to prevent DNS search domain issues
- Dynamically select the correct address type (IPv4/IPv6) when retrieving and validating endpoint slices
- Switch request pod image from **busybox** to **nicolaka/netshoot** for better IPv6 support and networking tools (`nc` with -4/-6 flags, `nslookup` with AAAA record support)
- For IPv6 connectivity tests, use explicit DNS resolution (`nslookup`) and pass the resolved IP to `nc`, as `nc -6` with hostname hangs in IPv6-only environments
- Validate that `ServiceImport` IPs match the service's `IPFamilies` in count and family type

Server-side changes:
- Create `socatListenerScript()` that auto-detects pod IP configuration using `MY_POD_IPS` environment variable
- Configure `socat` listeners appropriately:
  * IPv4-only: TCP-LISTEN
  * IPv6-only: TCP6-LISTEN
  * Dual-stack: TCP6-LISTEN with ipv6only=0

SRV record test improvements:
- Wait for expected number of SRV records across all IP families to be present before validating
- Collect endpoints for all IP families in dual-stack headless service tests
- Strip trailing periods from parsed SRV record domain names

Helper functions added:
- `dnsRecordTypeOf()`: derive DNS record type (A/AAAA) from IP family
- `addressTypeOf()`: derive `EndpointSlice` address type from IP family
- `ipFamilyOf()`: determine IP family from IP address string
- `ncCommand()`: generate `nc` command with IP family-specific handling
- `socatListenerScript()`: generate dual-stack aware socat listener script

### DIFF
--- a/conformance/conformance_suite.go
+++ b/conformance/conformance_suite.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	rest "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	k8snet "k8s.io/utils/net"
 	"sigs.k8s.io/mcs-api/pkg/apis/v1alpha1"
 	mcsclient "sigs.k8s.io/mcs-api/pkg/client/clientset/versioned"
 )
@@ -179,7 +180,7 @@ func newTestDriver() *testDriver {
 		}
 
 		// Set up the remote service (the first cluster is considered to be the remote)
-		t.deployHelloService(&clients[0], t.helloService)
+		t.helloService = t.deployHelloService(&clients[0], t.helloService)
 
 		// Start the request pod on all clusters
 		for _, client := range clients {
@@ -219,7 +220,7 @@ func (t *testDriver) deleteServiceExport(c *clusterClients) {
 	By(fmt.Sprintf("Service \"%s/%s\" unexported on cluster %q", t.namespace, helloServiceName, c.name))
 }
 
-func (t *testDriver) deployHelloService(c *clusterClients, service *corev1.Service) {
+func (t *testDriver) deployHelloService(c *clusterClients, service *corev1.Service) *corev1.Service {
 	if t.helloDeployment != nil {
 		_, err := c.k8s.AppsV1().Deployments(t.namespace).Create(ctx, t.helloDeployment, metav1.CreateOptions{})
 		Expect(err).ToNot(HaveOccurred())
@@ -228,7 +229,10 @@ func (t *testDriver) deployHelloService(c *clusterClients, service *corev1.Servi
 	deployed, err := c.k8s.CoreV1().Services(t.namespace).Create(ctx, service, metav1.CreateOptions{})
 	Expect(err).ToNot(HaveOccurred())
 
-	By(fmt.Sprintf("Service \"%s/%s\" deployed on cluster %q", deployed.Namespace, deployed.Name, c.name))
+	By(fmt.Sprintf("Service \"%s/%s\" with IP families %v deployed on cluster %q", deployed.Namespace, deployed.Name,
+		deployed.Spec.IPFamilies, c.name))
+
+	return deployed
 }
 
 func (t *testDriver) getServiceImport(c *clusterClients, name string) *v1alpha1.ServiceImport {
@@ -246,6 +250,8 @@ func (t *testDriver) getServiceImport(c *clusterClients, name string) *v1alpha1.
 func (t *testDriver) awaitServiceImport(c *clusterClients, name string, reportNonConformanceOnMissing bool,
 	verify func(Gomega, *v1alpha1.ServiceImport)) *v1alpha1.ServiceImport {
 	var serviceImport *v1alpha1.ServiceImport
+
+	By(fmt.Sprintf("Retrieving ServiceImport for %q on cluster %q", name, c.name))
 
 	Eventually(func(g Gomega) {
 		si := t.getServiceImport(c, name)
@@ -270,6 +276,16 @@ func (t *testDriver) awaitServiceImport(c *clusterClients, name string, reportNo
 	}).Within(20 * time.Second).WithPolling(100 * time.Millisecond).Should(Succeed())
 
 	return serviceImport
+}
+
+func (t *testDriver) awaitServiceImportIPFamilies(c *clusterClients) []corev1.IPFamily {
+	serviceImport := t.awaitServiceImport(c, t.helloService.Name, false,
+		func(g Gomega, serviceImport *v1alpha1.ServiceImport) {
+			g.Expect(serviceImport.Spec.IPFamilies).NotTo(BeEmpty(),
+				"ServiceImport on cluster %q does not contain an IP family", c.name)
+		})
+
+	return serviceImport.Spec.IPFamilies
 }
 
 func (t *testDriver) awaitNoServiceImport(c *clusterClients, name, nonConformanceMsg string) {
@@ -373,13 +389,15 @@ func (t *testDriver) awaitServicePodIP(c *clusterClients) string {
 }
 
 func (t *testDriver) execPortConnectivityCommand(port int, matchStr string, nIter int) {
-	command := []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.%s %d",
-		t.helloService.Name, t.namespace, dnsDomain, port)}
-
 	for _, client := range clients {
-		By(fmt.Sprintf("Executing command %q on cluster %q", strings.Join(command, " "), client.name))
+		for _, ipFamily := range t.awaitServiceImportIPFamilies(&client) {
+			serviceFQDN := fmt.Sprintf("%s.%s.svc.%s", t.helloService.Name, t.namespace, dnsDomain)
+			command := []string{"sh", "-c", ncCommand(ipFamily, serviceFQDN, port)}
 
-		t.awaitCmdOutputMatches(&client, command, matchStr, nIter, reportNonConformant(""))
+			By(fmt.Sprintf("Executing %s command %q on cluster %q", ipFamily, strings.Join(command, " "), client.name))
+
+			t.awaitCmdOutputMatches(&client, command, matchStr, nIter, reportNonConformant(""))
+		}
 	}
 }
 
@@ -448,4 +466,53 @@ func requireTwoClusters() {
 	if len(clients) < 2 {
 		Skip("This test requires at least 2 clusters - skipping")
 	}
+}
+
+func addressTypeOf(f corev1.IPFamily) discoveryv1.AddressType {
+	if f == corev1.IPv4Protocol {
+		return discoveryv1.AddressTypeIPv4
+	}
+
+	return discoveryv1.AddressTypeIPv6
+}
+
+func dnsRecordTypeOf(f corev1.IPFamily) string {
+	if f == corev1.IPv4Protocol {
+		return "A"
+	}
+
+	return "AAAA"
+}
+
+func ipFamilyOf(ip string) corev1.IPFamily {
+	f := k8snet.IPFamilyOfString(ip)
+	Expect(f).NotTo(Equal(k8snet.IPFamilyUnknown))
+
+	if f == k8snet.IPv4 {
+		return corev1.IPv4Protocol
+	}
+
+	return corev1.IPv6Protocol
+}
+
+// ncCommand creates a shell command that connects to a service using nc with the specified IP family.
+// The netshoot image provides nc which supports -4/-6 flags for forcing IP family.
+func ncCommand(ipFamily corev1.IPFamily, serviceFQDN string, port int) string {
+	ipFlag := "-4"
+	if ipFamily == corev1.IPv6Protocol {
+		ipFlag = "-6"
+	}
+
+	// For IPv6, nc -6 with hostname seems to hang in IPv6-only environments
+	// Fall back to explicit DNS resolution and pass IP directly to nc
+	if ipFamily == corev1.IPv6Protocol {
+		fqdnWithDot := serviceFQDN + "."
+		return fmt.Sprintf(
+			"addr=$(nslookup -type=AAAA %s 2>&1 | grep '^Address:' | grep -v '#' | tail -1 | awk '{print $2}'); "+
+				"if [ -z \"$addr\" ]; then echo 'No IPv6 address found'; exit 1; fi; "+
+				"echo hi | nc -v -w 2 $addr %d 2>&1",
+			fqdnWithDot, port)
+	}
+
+	return fmt.Sprintf("echo hi | nc -v -w 2 %s %s %d 2>&1", ipFlag, serviceFQDN, port)
 }
--- a/conformance/connectivity.go
+++ b/conformance/connectivity.go
@@ -37,14 +37,18 @@ var _ = Describe("", func() {
 			AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#exporting-services")
 			By("attempting to access the remote service", func() {
 				By("issuing a request from all clusters", func() {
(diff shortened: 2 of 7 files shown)


tpantelis:
@MrFreezeex Please try out with Cilium when you get a chance.

MrFreezeex:
Hey @tpantelis, I can confirm that Cilium works with this, I haven't yet got time to fully review it  but at first glance this looks great, thanks a lot!

conformance/k8s_objects.go
```diff
@@ -67,18 +68,59 @@ func newHelloServiceExport() *v1alpha1.ServiceExport {
...
 }
 
+// socatListenerScript generates a shell script that detects the pod's IP family configuration
+// and starts the appropriate socat listener (IPv4, IPv6, or dual-stack).
+func socatListenerScript(protocol string) string {
+	return `# Detect IP families available on the pod from podIPs
+has_ipv4=false
+has_ipv6=false
+if echo "$MY_POD_IPS" | grep -q ':'; then has_ipv6=true; fi
+if echo "$MY_POD_IPS" | grep -v ':' | grep -q '[0-9]'; then has_ipv4=true; fi
```
MrFreezeex: Does MY_POD_IPs always on multiple lines? If it's IP appended on one line it look like this won't work because of the `grep -v` that would discard the whole line 🤔

conformance/k8s_objects.go
```diff
@@ -67,18 +68,59 @@ func newHelloServiceExport() *v1alpha1.ServiceExport {
...
+
+# Debug: log what we detected
+echo "DEBUG: MY_POD_IPS=$MY_POD_IPS" >&2
+echo "DEBUG: has_ipv4=$has_ipv4, has_ipv6=$has_ipv6" >&2
+
+# Get first IP for echoing back - podIPs might be JSON array or space-separated
```
MrFreezeex: Can it really have multiple formats :O?

conformance/endpoint_slice.go
```diff
@@ -41,27 +41,29 @@ var _ = Describe("", Label(OptionalLabel, EndpointSliceLabel), func() {
...
-							discoveryv1.LabelManagedBy, K8sEndpointSliceManagedByName)))
+				for _, ipFamily := range t.helloService.Spec.IPFamilies {
+					eps := t.awaitMCSEndpointSlice(&client, addressTypeOf(ipFamily), nil, reportNonConformant(fmt.Sprintf(
+						"an MCS EndpointSlice was not found on cluster %q. An MCS EndpointSlice is identified by the presence "+
+							"of the required MCS labels (%q and %q). "+
+							"If the MCS implementation does not use MCS EndpointSlices, you can specify a Ginkgo label filter using "+
+							"the %q label where appropriate to skip this test.",
+						client.name, v1alpha1.LabelServiceName, v1alpha1.LabelSourceCluster, EndpointSliceLabel)))
+
+					endpointSlices[i] = eps
```
MrFreezeex: This would override previously defined endpointslice when you have multiple ip families, I would suggest to use append to add to the slice here and change the make on line 41 to this style `make(make([]*discoveryv1.EndpointSlice, 0, xxx)`

conformance/endpoint_slice.go
```diff
@@ -41,27 +41,29 @@ var _ = Describe("", Label(OptionalLabel, EndpointSliceLabel), func() {
...
-				Expect(eps.Labels).To(HaveKey(discoveryv1.LabelManagedBy),
-					reportNonConformant(fmt.Sprintf("the MCS EndpointSlice %q does not contain the %q label",
-						eps.Name, discoveryv1.LabelManagedBy)))
-
-				if !skipVerifyEndpointSliceManagedBy {
-					Expect(eps.Labels[discoveryv1.LabelManagedBy]).ToNot(Equal(K8sEndpointSliceManagedByName),
-						reportNonConformant(fmt.Sprintf("the MCS EndpointSlice's %q label must not reference %q",
-							discoveryv1.LabelManagedBy, K8sEndpointSliceManagedByName)))
+				for _, ipFamily := range t.helloService.Spec.IPFamilies {
+					eps := t.awaitMCSEndpointSlice(&client, addressTypeOf(ipFamily), nil, reportNonConformant(fmt.Sprintf(
```
MrFreezeex: Hmm ServiceImport IPFamilies could be potentially different from the one of the exported Service so this could break on conformant implementation

conformance/service_import.go
```diff
@@ -202,16 +202,24 @@ func testClusterIPServiceImport() {
...
 		Label(RequiredLabel), func() {
 			serviceImport := t.awaitServiceImport(&clients[0], t.helloService.Name, false,
 				func(g Gomega, serviceImport *v1alpha1.ServiceImport) {
 					g.Expect(serviceImport.Spec.IPs).ToNot(BeEmpty(), reportNonConformant(""))
 				})
 
-			Expect(net.ParseIP(serviceImport.Spec.IPs[0])).ToNot(BeNil(),
-				reportNonConformant(fmt.Sprintf("The value %q is not a valid IP", serviceImport.Spec.IPs[0])))
+			Expect(serviceImport.Spec.IPs).To(HaveLen(len(t.helloService.Spec.IPFamilies)), reportNonConformant(
+				"The ServiceImport IPs field must have the same number of IPs as the corresponding Service's IPFamilies"))
```
MrFreezeex: ServiceImport IPFamilies are entirely implementation dependent so this isn't necessarily true for a conformant implementation.

We could potentially check if the ServiceImport IPs are in the same order as the ServiceImport IPFamilies but I think that would be it 🤔

conformance/headless_service_dns.go
```diff
@@ -49,19 +49,23 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, HeadlessLabel), func() {
 		"ready endpoint addresses of all the backing pods", func() {
 		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns")
 
-		command := []string{"sh", "-c", fmt.Sprintf("nslookup %s.%s.svc.%s", t.helloService.Name, t.namespace, dnsDomain)}
+		for _, ipFamily := range t.helloService.Spec.IPFamilies {
```
MrFreezeex: Similarly as some other comments we probably need to iterate on ServiceImport IPFamilies here instead

conformance/headless_service_dns.go
```diff
@@ -80,35 +84,39 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, HeadlessLabel), func() {
 			AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns")
 
 			for _, client := range clients {
-				eps := t.awaitMCSEndpointSlice(&client, discovery.AddressTypeIPv4, func(g Gomega, eps *discovery.EndpointSlice) {
-					g.Expect(eps.Endpoints).To(HaveLen(replicas),
-						"the MCS EndpointSlice %q does not contain the expected number of endpoints %d",
-						eps.Name, replicas)
+				for _, ipFamily := range t.helloService.Spec.IPFamilies {
```
MrFreezeex: Same here for the IPFamilies

conformance/headless_service_dns.go
```diff
@@ -118,17 +126,20 @@ var _ = Describe("", Label(OptionalLabel, DNSLabel, HeadlessLabel), func() {
 		"records", func() {
 		AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#dns")
 
-		endpoints := t.awaitK8sEndpoints(&clients[0], discovery.AddressTypeIPv4)
+		// Collect endpoints for all IP families in the service (for dual-stack support)
+		var allEndpoints []endpointInfo
+		for _, ipFamily := range t.helloService.Spec.IPFamilies {
```
MrFreezeex, line 130: Same here for the IPFamilies

conformance/conformance_suite.go
```diff
@@ -373,13 +377,15 @@ func (t *testDriver) awaitServicePodIP(c *clusterClients) string {
 }
 
 func (t *testDriver) execPortConnectivityCommand(port int, matchStr string, nIter int) {
-	command := []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.%s %d",
-		t.helloService.Name, t.namespace, dnsDomain, port)}
+	for _, ipFamily := range t.helloService.Spec.IPFamilies {
```
MrFreezeex: Same here for IPFamilies

MrFreezeex (changes requested):

conformance/k8s_objects.go
tpantelis: TBH, I let `claude` write this part 😄 It assumed the format could be plain IPs or json-formatted - I didn't question it. But on further investigation, it should be simple comma-separated IPs. I've tested this in dual-stack and it works so the `grep -v` didn't come into play. I simplified it.

conformance/k8s_objects.go
tpantelis: No - I think `claude` hallucinated a bit 😄

conformance/headless_service_dns.go
tpantelis: Looking at the spec:  "_An implementation may use the ipFamilies field from the exported Services as a hint to influence the IPs and ipFamilies of the ServiceImport object. The exact mechanism for determining those fields is implementation-defined._ "

This is a bit nebulous. I'm not sure if it means that an implementation isn't required to set `IPFamilies` on `ServiceImport`. But if we want to make the assumption that implementations will set it to something then I think that's reasonable (note that Submariner currently doesn't but there's an issue filed for it so we intend to). We need to rely on something basic in order to provide conformance tests. If `ServiceImport.IPFamilies` isn't set then we could fall back to the exported `Service.IPFamilies`.

conformance/endpoint_slice.go
tpantelis: Can you elaborate? Do you mean a case where the service is exported on cluster1 which is dual-stack and imported on cluster2 which is IPv4-only? If so then in that case cluster2 would fail due to missing IPv6 EPS. But before these changes, a similar scenario could occur as well. Previously we assumed IPv4 and looked for an `discoveryv1.AddressTypeIPv4` EPS. So if cluster2 was IPv6-only, it would've failed.

We could get elaborate and check which IP families are supported on the other importing clusters. Perhaps we could do that as a future enhancement. For now at least I think it's reasonable to assume the clusters are homogenous wrt IP family support.

tpantelis: Thinking further on this, if we assume that we can rely on `ServiceImport.IPFamilies` being set properly, then we could retrieve the `ServiceImport` for each cluster and use its `IPFamilies` to check for the appropriate EPS's.

conformance/endpoint_slice.go
tpantelis: yes - good catch. We actually need a multi-dimensional array.

conformance/headless_service_dns.go
MrFreezeex: Hmm that's true indeed, maybe we could update the KEP to make IPFamilies required while its content should still be implementation dependent? That would make it a field that we could rely on for anything ready ServiceImport including the conformance test?

tpantelis: yeah I think it makes sense for `IPFamilies` to be required and match `IPs`, same as `Service`.  But for now and since `IPFamilies` is newer, for compatibility I think we should use `Service.IPFamilies` if `ServiceImport.IPFamilies` isn't set. WDYT?

conformance/service_import.go
tpantelis: I changed it to verify `ServiceImport.IPs` are in the same order as the `ServiceImport.IPFamilies` (if set).

conformance/headless_service_dns.go
MrFreezeex: Well there is no direct link between the two enforced by the KEP, you could have `Service.IPFamilies=IPv4` while `ServiceImport.IPFamilies=IPv6` technically :/. And more realistically in Cilium we are doing an intersection of all IPFamilies exported so if I were to launch the conformance test with one cluster with, let's say, dual stack and the other in IPv4, there are probably some test here that that would fail :sweat_smile:.

tpantelis: Well `Service.IPFamilies=IPv4` while `ServiceImport.IPFamilies=IPv6` wouldn't make sense. 

So it sounds like using `ServiceImport.IPFamilies`, as you initially suggested, is the reasonable approach (that should handle Cilium's intersection approach). But I'll fall back to `Service.IPFamilies` if `ServiceImport.IPFamilies` isn't set (need something to go on).

MrFreezeex: Sounds good to me!

tpantelis:
> Hey @tpantelis, I can confirm that Cilium works with this, I haven't yet got time to fully review it but at first glance this looks great, thanks a lot!

I addressed all your comments. I have modified the DNS/connectivity related tests to obtain the IP families from the `ServiceImport` spec as discussed (I updated Submariner to populate `ServiceImport.Spec.IPFamilies` and everything passes).

conformance/connectivity.go
```diff
@@ -37,8 +37,8 @@ var _ = Describe("", func() {
 			AddReportEntry(SpecRefReportEntry, "https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#exporting-services")
 			By("attempting to access the remote service", func() {
 				By("issuing a request from all clusters", func() {
-					command := []string{"sh", "-c", fmt.Sprintf("echo hi | nc %s.%s.svc.%s 42",
-						t.helloService.Name, t.namespace, dnsDomain)}
+					serviceFQDN := fmt.Sprintf("%s.%s.svc.%s", t.helloService.Name, t.namespace, dnsDomain)
+					command := []string{"sh", "-c", ncCommand(t.helloService.Spec.IPFamilies[0], serviceFQDN, 42)}
```
MrFreezeex: nit: maybe we could run this on all IPFamilies for completness?

conformance/k8s_objects.go
```diff
@@ -164,7 +216,7 @@ func newRequestPod() *corev1.Pod {
 			Containers: []corev1.Container{
 				{
 					Name:  "request",
-					Image: "busybox",
+					Image: "nicolaka/netshoot:latest",
```
MrFreezeex: I reckon it was the same as before but I am wondering if we shouldn't pin the image to like `v0.15` instead of `latest` so that ImagePullPolicy would implicitly be "IfNotPresent" and potentially minimize some docker hub rate limit issues 🤔 . And also avoid potentially future breaking change and allow running old conformance test versions without issues here.

MrFreezeex (approved):
Sorry for the wait, lgtm overall I left two minor questions/suggestions inline

conformance/connectivity.go
tpantelis: done

conformance/k8s_objects.go
tpantelis: done

MrFreezeex (approved):
Thanks! :D 
/lgtm

skitt:
Based on the comments, it seems Claude was used to develop this, which should be mentioned in the PR description (*not* the commit message); see https://github.com/kubernetes/community/blob/main/contributors/guide/pull-requests.md#ai-guidance (for future reference).

skitt:
/approve

Thanks!

PR #145, opened 2026-03-16
conformance: generate report in yaml format

### LINKED ISSUE #144: Support machine-readable output for conformance test reports
Currently, the MCS API conformance testing tool generates an HTML report. HTML is good for human readability but having a machine readable output like a YAML or JSON makes it easy it easy to create the comparison matrix among all the implementations.

Ginkgo (the framework used in the conformance tests) can produce a JSON report, perhaps that can be used directly here.

AnupamGhosh:
## Summary 
Fixes #144
Generate report in yaml format from the conformance reporting tool along with the html report. 

The yaml report also includes `total` (total number of tests in the suite) and `passed` (number of tests that passed) fields for a quick summary of conformance results.

Below is how the yaml file looks
```yaml
groups:
    - name: Required
      tests:
...
        - desc: Connectivity to a service that is not exported should be inaccessible
          ref: https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#exporting-services
          labels: []
          passed: true
          failed: false
          skipped: false
          conformant: true
          message: ""
...
suitefailure: ""
dnsdomain: clusterset.local
passed: 2
total: 27
```

### DIFF
--- a/conformance/report.go
+++ b/conformance/report.go
@@ -31,6 +31,7 @@ import (
 	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/matchers"
+	"gopkg.in/yaml.v3"
 )
 
 const (
@@ -60,6 +61,7 @@ type testInfo struct {
 	Desc       string
 	Ref        string
 	Labels     []string
+	Passed     bool
 	Failed     bool
 	Skipped    bool
 	Conformant bool
@@ -166,7 +168,7 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
 			}
 
 			info := testInfo{
-				Desc:       specReport.FullText(),
+				Desc:       strings.TrimSpace(specReport.FullText()),
 				Conformant: true,
 			}
 
@@ -202,6 +204,8 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
 				info.Message = parseFailureMessage(specReport.FailureMessage())
 			}
 
+			info.Passed = !info.Failed && !info.Skipped && info.Conformant
+
 			if info.Message != "" {
 				info.Message = " - " + info.Message
 			}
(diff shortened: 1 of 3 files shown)


MrFreezeex:
Hi :wave: so did you tested the default JSON output from ginkgo?

anupamdialpad:
Hi @MrFreezeex , yes I had tested with the native json reporting option of ginko with 
```
go -C conformance run github.com/onsi/ginkgo/v2/ginkgo --json-report=report.json . \
      -- --kubeconfig=~/.kube/config --contexts=kind-c1,kind-c2
```
report.json contains all the raw data of the test result (the json file has >8k lines). The result is not structured like report.html based on the struct `data` in `ReportAfterSuite`.

To have a structured output like report.html I used the approach in this PR to generate report.yaml file.

conformance/report.go
```diff
@@ -202,6 +204,8 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
 				info.Message = parseFailureMessage(specReport.FailureMessage())
 			}
 
+			info.Passed = !info.Failed && !info.Skipped && info.Conformant
```
MrFreezeex, line 207: Is it actually possible to have conformant true while failed or skipped is also true? Said differently does it matter to add this new boolean or it's actually the same as relying on the existing conformant boolean?

MrFreezeex (commented):
Thanks and sorry for the delay! I left one quick comment/question inline but the PR makes sense to me overall!

Side note, technically the total tests and number of tests succeeded is a separate feature. It looks good but for the future adding a quick note that you are adding this new feature in the PR description would have made it even better! (Considering the PR size it's fine though!)

conformance/report.go
AnupamGhosh: I actually copied the logic from `report_template.gothml` where being conformant means test is "!skipped & !failed & conformant" and `info.Passed` means being conformant
https://github.com/kubernetes-sigs/mcs-api/blob/15301c2b6a81941dfd8be643f1719b6aa6104631/conformance/report_template.gohtml#L38-L43

Also looking at the code below there is a case where both `info.Conformant` & `info.failed` can be true
https://github.com/kubernetes-sigs/mcs-api/blob/15301c2b6a81941dfd8be643f1719b6aa6104631/conformance/report.go#L197-L201

AnupamGhosh:
Thanks @MrFreezeex for the feedback! I have edited the description of the PR to include that total & passed tests are added as part of this PR

MrFreezeex (approved):
Thanks for the additional info!
/lgtm

skitt:
/approve

Thanks!

PR #150, opened 2026-04-25
Conformance: Track required tests in report.yaml

AnupamGhosh:
### Summary
Include the total no. of required tests and the no. of required tests an implementation passed. The idea came from https://github.com/kubernetes-sigs/sig-multicluster-site/pull/54 where the conformance matrix need to show if an implementation passed all the required tests which determines if an implementation is conformant or not. 

### Test Plan
1. Running the conformance tool showed that report.yaml included two new fields `requiredpassed` and `requiredtotal`
```yaml
dnsdomain: clusterset.local
passed: 2
total: 27
requiredpassed: 1
requiredtotal: 14
```

### DIFF
--- a/conformance/report.go
+++ b/conformance/report.go
@@ -236,8 +236,19 @@ var _ = ReportAfterSuite("MCS conformance report", func(report Report) {
 
 	totalTests := 0
 	passedTests := 0
+	requiredTotal := 0
+	requiredPassed := 0
 	for _, g := range testGroups {
 		for _, t := range g.Tests {
+			// Track required tests. Implementation is conformant if all required tests pass.
+			if g.Name == RequiredLabel {
+				requiredTotal++
+				if t.Passed {
+					requiredPassed++
+				}
+			}
+
+			// Overall totals include both required and optional tests.
 			totalTests++
 			if t.Passed {
 				passedTests++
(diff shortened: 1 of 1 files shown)


MrFreezeex:
/ok-to-test

Hi @AnupamGhosh :wave:, thanks for the PR!

I am wondering if we need to include that info in the yaml... Like with your python script that generate the markdown matrix, it seems easy enough to distinguish that info using the labels. And if we were to add those counter for the required tests we could be tempted to add some other counters in the future for the other categories/labels too which kind of unnecessarily duplicate that info :thinking:.

MrFreezeex:
I just realized that the optional/required labels were not directly passed but it's still separated into groups in the report so most of what I previously wrote kind of still works ~.

AnupamGhosh:
Hi @MrFreezeex, thanks for reviewing the PR!

I was thinking from another aspect
1. To figure out whether an implementation is conformat all the required tests need to iterated to check if they passed. To avoid this computation (even though small) everytime during site build I thought of adding it in report.yaml file which can be fetched directly by the [python script](https://github.com/AnupamGhosh/sig-multicluster-site/blob/15d2b70dea39f9bd9f1fcf6ebcadd47269e6bb82/hack/generate-conformance-matrix.py) that generates the comparison matrix.
2. I was trying to keep the python script as generic as possible so that other multicluster project can utilize this to generate the comparison matrix. If the format of report.yaml of other multicluster project is different then by adding these 4 fields -- `passed`, `total`, `requiredpassed` and `requiredtotal` in report.yaml the python script can easily generate the comparison matrix.

MrFreezeex:
Ah I see what you are trying to achieve now. I don't think any of our other sig-multicluster projects currently have a conformance tests though.

And overall the point of generating matrixes is to help people compare projects and at the same time many features are optional. So while getting the required test to pass and showing it is important it would likely be even more meaningful to compare the other optional tests either individually or by label which means that we would need to loop through the test arrays at some point anyway...

AnupamGhosh:
At this point I don't have a clear view of when the optional tests need to be iterated. But your point about developers getting tempted to add more additional fields to the report.yaml makes sense to me. So, I will move the logic to count the no. of required tests passed to the python script that generates the matrix. Thanks!

PR #151, opened 2026-04-29
Propagate the Ginkgo context

skitt:
Ginkgo provides its own context, using that instead of a shared context.TODO() provides better support for cancellation etc.


MrFreezeex (approved):
Looks great, thanks! I will lgtm and hold to let you decide if you want to let Tom have a look or just cancel the hold and merge.

/lgtm
/hold

PR #152, opened 2026-04-29
Reference gopkg.in/yaml.v3 directly

skitt:
It is used for reports.


MrFreezeex (approved):
Indeed thanks!

/lgtm

PR #154, opened 2026-05-06
conformance: increase timeout for pods start up

MrFreezeex:
In Cilium we are running this conformance test in CI and have seen about ~10-15% of failure and most of them could be attributed to a timeout waiting for a Pod to be ready.

Thus, this commit increase all occurrences in the conformance suite that actively wait for a pod to be ready (and related EndpointSlices) from 20s to 60s. This should let us know that that something is actually wrong with pod startup while not being too generous either.


MrFreezeex:
cc @tpantelis @skitt :eyes:

PR #157, opened 2026-06-18
conformance: add a short intro readme

MrFreezeex:
Since we are releasing a new version soon and that the conformance reports program is almost good to go I wrote this so that we won't have to potentially do a new release to only add some small readme like that.

My goal here is to just quickstart someone running the test and point it to the conformance report instruction, not necessarily be super exhaustive (at least at first, we can iterate on it later if needed)

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -1,0 +1,33 @@
+# MCS Conformance Tests
+
+This package contains the MCS API conformance suite. It is intended for MCS
+implementations to run against real clusters and generates `report.yaml` and
+`report.html` files. To produce valid results, the conformance suite needs to
+run against two or more clusters.
+
+## Run
+
+Install your MCS implementation and make sure your kubeconfig has contexts that
+can access the clusters you want to test against.
+
+Then run the tests with the following command, replacing `cluster-a` and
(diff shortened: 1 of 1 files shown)


MrFreezeex:
cc @AnupamGhosh

AnupamGhosh (approved):
This is nice! Adding the command on how to get the conformance report is helpful to end users!

conformance/README.md
```diff
@@ -0,0 +1,33 @@
+# MCS Conformance Tests
+
+This package contains the MCS API conformance suite. It is intended for MCS
+implementations to run against real clusters and generates `report.yaml` and
+`report.html` files. The conformance suite typically need to run against two
+or more clusters.
```
skitt:
```suggestion
`report.html` files. To produce valid results, the conformance suite needs to run against two
or more clusters.
```

skitt (commented):
Minor tweak to avoid uncertainty around “typically”.

conformance/README.md
MrFreezeex: I applied your suggestion, thanks!

skitt:
/lgtm
/approve

Thanks!

PR #161, opened 2026-07-24
Add multicluster category to CRDs

kahirokunn:
## Summary

Add the `multicluster` category to the MCS API CRDs.

## Background

Following [this discussion](https://github.com/kubernetes-sigs/cluster-inventory-api/pull/80#issuecomment-5071886576), SIG Multicluster CRDs should use a common category. This allows users to list these resources together with `kubectl get multicluster`.


skitt:
@MrFreezeex what do you think?

skitt:
From my perspective this is fine of course. The CRD version is bumped so this can go in as-is, I’ll approve it so it doesn’t have to wait for me if someone else LGTMs.

/approve

MrFreezeex (approved):
LGTM thank you (TIL that category was a thing 😅)
/lgtm

PR #165, opened 2026-09-01
Add kahirokunn as reviewer

kahirokunn:
Adding myself to the OWNERS file as discussed in the Sep-01-2026 SIG-MC call.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -5,6 +5,7 @@ reviewers:
   - lauralorenz
   - skitt
   - MrFreezeex
+  - kahirokunn
   - RainbowMango
   - ryanzhang-oss
   - tpantelis


MrFreezeex:
Heyy 👋, would be super glad to have you as a reviewer in the MCS-API repo given your previous contributions/involvements in the MCS-API area but I don't think this should come "built-in" with the new tech lead role.

Like what I mean is that ideally reviewers should IMO be supposed to actively reviewing PR for a given repo. AFAICT the difference of being reviewer or not is mostly about who get automatically assigned for reviews given that any kubernetes members should be able to add a lgtm if they want to.

We definitely need to discuss more about that in the next meeting (cc @skitt @lauralorenz @mikeshng or anyone else that wants to chime in) but separately if you do mean to actively review PRs from this repo specifically I would definitely support this PR! 🙏

mikeshng:
>  don't think this should come "built-in" with the new tech lead role.
> Like what I mean is that ideally reviewers should IMO be supposed to actively reviewing PR for a given repo.

+1 I completely agree.

skitt:
In the SIG call there was broad agreement that pursuing TL-style nominations for a few people including you was a good thing, however:
* we need to figure out what the TL role actually entails in SIG-MC (based on https://github.com/kubernetes/community/blob/main/contributors/chairs-and-techleads/technical-lead.md)
* member/reviewer/approver is separate, see https://github.com/kubernetes/community/blob/main/community-membership.md#reviewer for reviewers in particular

We do need to get our house in order so this will be a high-priority topic over the next few weeks, but let’s wait until things have settled down a little :wink:.

PR #167, opened 2026-09-04
Finish the Ginkgo v2 migration in the controllers suite

magic-peach:
TestControllers never called RunSpecs, so every Describe/It in this package, including the BeforeSuite/AfterSuite that provision a kind cluster, has silently run zero specs since the v1 to v2 migration dropped the old RunSpecsWithDefaultAndCustomReporters call. Once specs actually run, six It calls in serviceimport_test.go fail immediately with an unknown decorator error: a bare trailing int is Ginkgo v1's per-spec timeout syntax, which v2 does not accept. The Eventually calls inside each of those bodies already carry their own timeout, so the outer one was redundant even before it broke.

Tested: go build and go vet are clean. Ran the suite directly against a real kind cluster with KUBEBUILDER_ASSETS set, multiple times, to confirm both fixes are individually correct.

One thing worth flagging rather than papering over: once specs actually execute, the suite is flaky against a real cluster in a way unrelated to either fix above. One run hung on the EndpointSlice label sync check for its full timeout, another hit an immediate NoKindMatchError on ServiceImport creation right after CRD install. The differing symptom each run points at a real cluster timing or race issue rather than something this change introduces or could safely fix blind, and it may be specific to how CRD readiness and discovery behave in a sandboxed Docker environment rather than the project's own CI runners. Filing this PR scoped to the two confirmed syntax bugs and leaving that part for a maintainer to weigh in on.

/kind bug
/sig multicluster


tpantelis:
The controllers `BeforeSuite` setup fails in `clusterProvider.Create(clusterName)` trying to run `docker info --format '{{json .}}'`.  I don't know why but these are unit tests so I question why they need a real cluster setup. Looking at what the controllers and tests are doing, it seems to me the overhead and complexity of a real cluster is overkill and using a fake K8s client would suffice (especially testing `shouldIgnoreEndpointSlice` and `shouldIgnoreImport`). The E2E suite tests a real cluster.

magic-peach:
Thanks for looking into this, you are right that the setup was overkill. I dug into the actual CI failure and it is not just misconfiguration: the pull-mcs-api-test job has no Docker daemon at all, so clusterProvider.Create never had a chance to work there.

envtest does not actually need kind or a real cluster though: it can run its own etcd and kube-apiserver directly from the pinned binaries, no Docker involved. I dropped kind and UseExistingCluster and let envtest provision that control plane itself. This is a smaller change than switching to a fake client and keeps all 12 existing specs, including the ones that depend on the controller manager actually reconciling objects (shouldIgnoreEndpointSlice and shouldIgnoreImport are the only two that do not need any client at all).

Verified locally with the real envtest binaries and no Docker running:
- The `should be ignored` specs (pure function calls, the ones you mentioned) pass cleanly in under 6 seconds.
- A reconciliation spec (ServiceImport creating a derived Service) runs the full controller manager against the envtest API server successfully.

One thing worth flagging separately rather than silently: while verifying this I hit an intermittent `NoKindMatchError` for ServiceImport right after a fresh CRD install, in the reconciliation specs specifically. I can reproduce it locally, and it looks like a REST mapper discovery race rather than anything to do with kind vs plain envtest, since it is a client-side cache timing issue, not cluster-provisioning. I have not chased down a fix for that in this PR since it is orthogonal to the Docker problem and predates this change, but wanted to surface it rather than claim a fully clean run. Happy to open a follow-up issue for it if you agree it is worth tracking separately.

tpantelis:
> One thing worth flagging separately rather than silently: while verifying this I hit an intermittent `NoKindMatchError` for ServiceImport right after a fresh CRD install

I don't know where this is occurs but it seems like a timing issue as you mentioned, in which case I'd think retrying the operation (using `Eventually`) should alleviate it.

With your latest changes, it fails for another reason: `/usr/local/kubebuilder/bin/etcd: no such file or directory`.  For this to run in CI, it seems it would need to install/deploy the K8s control plane binaries (I don't have much experience using the envtest framework). Or simplify to use a fake client.

magic-peach:
You were right that it needed the control plane binaries, not a rewrite to a fake client. Pushed a fix: wired up sigs.k8s.io/controller-runtime/tools/setup-envtest the same way controller-gen is already invoked in this repo's tools module, pinned to 1.34 to match this repo's k8s.io/api version, and export KUBEBUILDER_ASSETS from it in the test target.

Verified locally with KUBEBUILDER_ASSETS unset beforehand: make test now fetches etcd/kube-apiserver on its own and the suite actually starts a real control plane instead of failing at BeforeSuite.

One side effect worth calling out directly: setup-envtest has no release that supports go 1.24, so this bumps tools/go.mod to go 1.26. That module has no source of its own outside tools.go's build-tag anchors used to pin dev tool versions, so it does not change what Go version the actual controller or CLI build against.

Also, running the full suite end to end with this change, I hit the same flaky EndpointSlice reconciliation timeout I flagged in the original PR description, this time it ran long enough to hit go test's own 10 minute timeout rather than failing fast. Still looks like a real cluster timing issue in this sandbox rather than anything caused by the envtest switch, but wanted to be upfront that it is not fully resolved.

tpantelis:
> Also, running the full suite end to end with this change, I hit the same flaky EndpointSlice reconciliation timeout I flagged in the original PR description, this time it ran long enough to hit go test's own 10 minute timeout rather than failing fast. Still looks like a real cluster timing issue in this sandbox rather than anything caused by the envtest switch, but wanted to be upfront that it is not fully resolved.

Can you describe the nature of the flaky EndpointSlice reconciliation? Does it look like an issue in the controller logic? But whatever the cause, we really shouldn't merge this PR with a known flaky test.

tpantelis:
It appears this PR also broke the E2E tests:

```
no required module provides package sigs.k8s.io/kind; to add it:
	go get sigs.k8s.io/kind
```

skitt:

/close

PR #172, opened 2026-09-07
Check scheme registration errors in servicecontroller

magic-peach:

`controllers/cmd/servicecontroller/servicecontroller.go`'s `init()` called `clientgoscheme.AddToScheme(scheme)` and `v1beta1.AddToScheme(scheme)` without checking the returned error. If scheme registration ever failed, the manager would start with a broken scheme and only fail later with a confusing "no kind registered" runtime error, instead of a clear startup failure at the actual point of the problem.

Checked both errors, following the same `setupLog.Error(err, ...)` + `os.Exit(1)` pattern already used in `main()` for the analogous `controllers.Start` error. Also switched `v1beta1.AddToScheme` to `v1beta1.Install`, since the generated `zz_generated.register.go` marks `AddToScheme` as `// Deprecated: use Install instead` (`Install` is the exact same underlying function, just the non-deprecated name).

Tested: `go build ./...` and `go vet ./...` pass in the `controllers` module. `golangci-lint run ./cmd/servicecontroller/...` (errcheck + staticcheck) is clean; it flagged both issues before this change.


controllers/cmd/servicecontroller/servicecontroller.go
```diff
@@ -37,8 +37,14 @@ var (
...
-	clientgoscheme.AddToScheme(scheme)
-	v1beta1.AddToScheme(scheme)
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		setupLog.Error(err, "problem registering client-go scheme")
+		os.Exit(1)
+	}
+	if err := v1beta1.Install(scheme); err != nil {
+		setupLog.Error(err, "problem registering v1beta1 scheme")
+		os.Exit(1)
+	}
```
skitt:
```suggestion
	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
	utilruntime.Must(v1beta1.Install(scheme))
```

(where `utilruntime` is `"k8s.io/apimachinery/pkg/util/runtime"`).

skitt (changes requested):

magic-peach:
@tpantelis  could you please merge this if no more changes are needed?

skitt:
Please squash the commits — essentially, keep only the changes from the second commit, with any parts of the commit message from the first commit that are useful for context.

magic-peach:
Squashed into a single commit.

skitt (approved):

skitt:

/close

PR #174, opened 2026-09-07
Remove dead RSS feed link from SECURITY.md

magic-peach:

`SECURITY.md`'s RSS-feed link 404s: Google Groups retired RSS feeds platform-wide, and there's no replacement URL. The current standard kubernetes-sigs SECURITY.md template (checked against kubernetes-sigs/kueue's live file) has already dropped this line, so this repo's copy is just stale.

Removed the sentence and the dead reference link, matching the current template exactly.

Note: this exact dead line is boilerplate copied into many kubernetes-sigs and kubernetes repos, not something unique to mcs-api, so feel free to point this at the shared template in kubernetes/community instead if that's a better place for it. Still a real, safe fix scoped to this repo's own file either way.

Doc-only change.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -4,8 +4,6 @@
 
 Join the [kubernetes-security-announce] group for security and vulnerability announcements.
 
-You can also subscribe to an RSS feed of the above using [this link][kubernetes-security-announce-rss].
-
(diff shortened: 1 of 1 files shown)


skitt:
/lgtm
/approve

Thanks!

PR #175, opened 2026-09-09
Return manager creation error instead of exiting process

magic-peach:

Start calls os.Exit(1) directly when ctrl.NewManager fails, even though every other error path in the same function returns the error to the caller. This makes the failure untestable and kills the whole process abruptly, which is especially disruptive for callers like the envtest suite that invoke Start in a goroutine.

Checked the production caller (controllers/cmd/servicecontroller/servicecontroller.go): it already checks the error returned by controllers.Start and calls os.Exit(1) itself, so production behavior is unchanged. This only fixes the internal inconsistency inside Start.

Tested: go build ./... and go vet ./... pass in both the root module and the controllers module.

### DIFF
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/base32"
+	"fmt"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -45,8 +46,7 @@ func derivedName(name types.NamespacedName) string {
 func Start(ctx context.Context, cfg *rest.Config, setupLog logr.Logger, opts ctrl.Options) error {
 	mgr, err := ctrl.NewManager(cfg, opts)
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
-		return err
+		return fmt.Errorf("unable to start manager: %w", err)
 	}
 
 	if err = (&ServiceImportReconciler{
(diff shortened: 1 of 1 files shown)


MrFreezeex (approved):
/ok-to-test
/lgtm

controllers/common.go
```diff
@@ -47,7 +46,7 @@ func Start(ctx context.Context, cfg *rest.Config, setupLog logr.Logger, opts ctr
 	mgr, err := ctrl.NewManager(cfg, opts)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
```
tpantelis: The caller also logs the error so we should wrap the return error with contextual info instead of logging here.

magic-peach:
Good point, fixed. Now wraps the error with context instead of logging it, since the caller already logs whatever Start returns.

controllers/common.go
```diff
@@ -46,8 +46,7 @@ func derivedName(name types.NamespacedName) string {
 func Start(ctx context.Context, cfg *rest.Config, setupLog logr.Logger, opts ctrl.Options) error {
 	mgr, err := ctrl.NewManager(cfg, opts)
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
-		os.Exit(1)
+		return fmt.Errorf("unable to start manager: %w", err)
```
tpantelis, line 49: Please do the same for the other errors below for consistency.

skitt:

/close

PR #176, opened 2026-09-11
Fix leftover template project name in RELEASE.md

magic-peach:
What changed: RELEASE.md still referred to this project as "The Kubernetes Template Project" and used kubernetes-template-project as the announcement email subject, both leftover from the kubernetes-sigs template repo this project was scaffolded from. Updated both to say Multi-Cluster Service APIs and mcs-api respectively.

Why: the release process doc should name the actual project, not the template it was generated from.

How tested: docs only change, read the diff for correctness.


skitt:

/close

PR #177, opened 2026-09-13
controllers: add test coverage for serviceImportOwner

magic-peach:

What changed: added test coverage for serviceImportOwner in controllers/service.go, covering no owner references, an unrelated owner, a ServiceImport reference with the wrong API version, a genuine ServiceImport owner, and a ServiceImport owner among other owner references. None of these were covered before.

Why: this function has zero prior test coverage. It uses the plain testing package rather than the envtest/kind based suite, so it does not depend on the ongoing Ginkgo v2 migration in PR #167.

How tested: go build ./..., go vet ./..., go test ./... in the controllers module. All 5 cases pass.


MrFreezeex (changes requested):
Hi 👋, the controllers package don't need to be particularly robust so I don't think we need to have additional testings beside the e2e tests

/close

PR #178, opened 2026-09-14
controllers: add test coverage for derivedName

magic-peach:

What changed: added common_test.go with a direct test for derivedName in controllers/common.go, covering determinism for the same input, different derived names for different inputs (both name and namespace variation), and the expected output format.

Why: derivedName had no direct test, only indirect use as a helper inside other tests to compute expected values.

How tested: gofmt -l, go vet ./..., go test ./... -run TestDerivedName in the controllers module. All checks pass.


MrFreezeex:
Similarly as https://github.com/kubernetes-sigs/mcs-api/pull/177#pullrequestreview-5196310572, I am closing this.

/close

PR #180, opened 2026-09-14
conformance: add test coverage for report parsing helpers

magic-peach:

What changed: added report_helpers_test.go with direct tests for firstLine, lookupSpecRef and parseFailureMessage in conformance/report.go.

Why: these pure string processing helpers back the conformance report generator and had no tests. parseFailureMessage's regex is built at runtime from the actual gomega HaveOccurredMatcher and SucceedMatcher output, so I probed that real output in a throwaway test first to get the assertions right rather than guessing the format.

How tested: gofmt -l, go vet ./..., go test ./... -run "TestFirstLine|TestLookupSpecRef|TestParseFailureMessage" in the conformance module. All 7 cases pass.


skitt:

/close

Issue #21, opened 2022-11-17
Intro tutorials for MCS API use cases


lauralorenz:
Develop some vendor agnostic intro tutorials about the MCS API that can be easily digested as part of a generic documentation push. Keep them implementation agnostic but I think it would be useful to tie them to some common use cases as suggested below.

Ideas:
* MCS for a single-cluster stateful service serving multiple clusters (single ServiceExport case) (see [this walkthrough](https://cloud.google.com/kubernetes-engine/docs/how-to/migrate-gke-multi-cluster#migrate_to_multi-cluster_services) for some inspo)
* MCS for a same-named stateless service in clusters in different regions for simple failover (merged ServiceExport case)
* MCS for a same-named stateless service for blue-green upgrade (see [this video](https://youtu.be/hztXW7Hk6Vo) for example)

holgerson97:
Hi @lauralorenz 

I am new to the project and would be happy to take over this part. First I wanted to clarify if this is something that is still needed, let me know what you think!

Kind regards,
Nils

levenhagen:
Hi MCS team. Is this still needed? @skitt

skitt:
@levenhagen it is indeed still needed!

aryansk:
I’m preparing a focused documentation PR with three vendor-neutral tutorials covering a single exported stateful service, regional stateless-service aggregation, and blue-green upgrades. I’ll open it after local validation.

PR #181, opened 2026-09-14
Add a multi region failover tutorial

magic-peach:
What changed: added docs/tutorials/multi-region-failover.md, a vendor agnostic tutorial walking through the merged ServiceExport case, exporting the same Service from clusters in different regions for automatic failover, and linked it from the README.

Why: this addresses part of issue #21. The existing demo only exercises the low level EndpointSlice mechanics of the API rather than the everyday user workflow, there is no example anywhere in the repo that actually shows creating a ServiceExport. This tutorial shows what a user creates and how failover behaves.

How tested: docs only change. Cross checked every technical claim against the current code before writing it: the API version used (v1beta1, the storage version per pkg/apis/v1beta1), the clusterset.local DNS suffix (confirmed as the spec default in conformance/conformance_suite.go), and that ServiceExport requires no spec fields to opt in (pkg/apis/v1beta1/serviceexport.go).


skitt:

/close

PR #182, opened 2026-09-20
Ignore a missing owning ServiceImport in the Service reconciler

magic-peach:

The Service reconciler (controllers/service.go) looks up its owning ServiceImport to sync the derived Service's ClusterIPs back onto it, but returned any error from that Get call including NotFound. Every other Get call in this package (endpointslice.go, serviceimport.go) already ignores NotFound with client.IgnoreNotFound. If the owning ServiceImport is deleted before Kubernetes garbage collection removes the Service it owns, this reconciler would return the NotFound error and controller-runtime would retry indefinitely on a resource that will never exist again, instead of returning cleanly like the rest of the package does in the equivalent situation.

This file and the serviceImportOwner helper had no tests at all. Added controllers/service_test.go: a plain unit test for serviceImportOwner, and a fake-client-based test for Reconcile that reproduces the missing-ServiceImport case directly (the full Ginkgo suite in this package needs a live kind cluster, not available here, so this uses sigs.k8s.io/controller-runtime/pkg/client/fake instead, following the same pattern already used for the pure helper functions elsewhere in this repo).

Tested: go test ./... in controllers/ passes for the new tests (go test ./... -run TestServiceReconcileMissingServiceImport|TestServiceImportOwner). Verified the bug is real by stashing just the fix in service.go and confirming TestServiceReconcileMissingServiceImport fails with the exact NotFound error before restoring the fix. go vet and gofmt are clean.


skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

PR #183, opened 2026-09-20
Check scheme registration errors in servicecontroller

magic-peach:

Both AddToScheme calls in controllers/cmd/servicecontroller/servicecontroller.go's init() had their error return ignored. If registration ever failed, the manager would start with a broken scheme and fail later with a confusing unregistered kind error instead of a clear startup failure at the actual point of the problem.

Wrapped both calls in utilruntime.Must, matching the pattern used elsewhere in the codebase for scheme registration. Also switched to v1beta1.Install, since v1beta1.AddToScheme is deprecated in favor of it.

Tested: go vet and go build both clean on controllers/cmd/servicecontroller. This is init-time registration of a static, always-valid scheme, so there is no failure path to exercise with a test; the change only affects behavior if that ever changes.


skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

